Issue #18, opened 2016-08-29
Support arrow functions


TatriX:
There is no support for syntax highlighting of the arrow function formal parameters, `beginning-of-defun` etc.

lddubeau:
I'm closing this due to the lack of an actual case illustrating the problem. I'll gladly reopen once a case other people can use to reproduce the problem is added to this issue. I use `typescript-mode` day in and day out and don't recall a specific problem with highlighting arrow functions. Maybe it's been fixed while fixing another problem... hard to tell without a case.

PR #19, opened 2016-08-29
Added basic syntax highlighting for arrow functions

TatriX:
Fast and simple partial sollution for #18 
There is a `"\\_<function\\_>"` pattern for matching functions. I think it should be somehow changed to match arrow functions as well.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -1543,8 +1543,12 @@ point of view of font-lock.  It applies highlighting directly with
 
     ;; arrow function formal parameters
     ,(list
-      "(\\(.*\\))\\s-*=>"
+      "(\\([^(]*?\\))\\s-*=>"
       '(1 font-lock-variable-name-face))
+
+    ,(list
+      (concat "\\(" typescript--name-re "\\)\\_>(")
+      '(1 font-lock-function-name-face))
 
     ,(list
       (concat "\\(" typescript--name-re "\\)\\s-*=>")
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -265,6 +265,12 @@ Match group 1 is the name of the function.")
   "Regexp matching a line in the typescript form \"var MUMBLE = function\".
 Match group 1 is MUMBLE.")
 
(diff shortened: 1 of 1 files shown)


ananthakumaran:
screen shot 2016-08-31 at 8 43 01 pm
screen shot 2016-08-31 at 8 42 42 pm

I did some quick testing, the regex seems to be too greedy. Particularly, I don't think using `.*` is such a good idea.

TatriX:
I've also added highlighting of the function call as in js2-mode but I didn't find a good default font-lock name so I've used `font-lock-function-name-face`.

josteink:
Sorry about the slow response. I'm trying this out, but I cannot reproduce that I'm getting a different face for the fat arrow (only for the function parameters).

Is it my theme which has weird values, or am I missing something obvious? :)

TatriX:
It's hard to say now ;) I think you can ignore this pr if it has flaws because I cant work on it now.

josteink:
I'll do some more testing without my custom Emacs theme being loaded and if that works, I'll see about merging this. 

Anyway thank you for your contribution!

josteink:
After testing this some more, I think we'll just have to drop it for now.

I'm getting different fontification for the following scenarios:

````
// entire param-list (commas included) fontified with highlight-symbol-face 
var x = (a, b, c) 
    //etc
};

// commas NOT fontified with highlight-symbol-face
function x(a, b, c) {
    //etc
};

// Not to mention type-annotations

// Variables highlighted with default
// Types with font-lock-keyword-face
var x = (a: string, b: int, c: boolean) => {
}

// Variables highlighted with highlight-symbol-face
// Types with font-lock-keyword-face
function x(a: string, b: int, c: boolean) {
    
}
````

All in all there's lackings with the current fontification without your patch too, but right now this patch adds to the inconsistency rather than reducing it.

Therefore I vote closing it. Hopefully nobody's feelings are trampled :)

TatriX:
Sure, good choice.

PR #24, opened 2016-11-21
Fix wrong cursor position in `typescript-indent-line`

hekt:
Cursor jumps to wrong position when editing line includes tab or [fullwidth character](https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms). Because `(current-column)` counts there as 2 or more characters.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -1769,7 +1769,7 @@ nil."
     (widen)
     (let* ((parse-status
             (save-excursion (syntax-ppss (point-at-bol))))
-           (offset (- (current-column) (current-indentation))))
+           (offset (- (point) (line-beginning-position) (current-indentation))))
       (indent-line-to (typescript--proper-indentation parse-status))
       (when (> offset 0) (forward-char offset)))))
 


josteink:
I'm willing to merge this if I can reproduce the error and verify the fix.

Do you have a sample TypeScript-file I can use to test with?

hekt:
```typescript
let cjk = 'あいう'
let tab = '	'

something();
```

type `;` at end of each line, the cursor will move to next line.

josteink:
Can reproduce. Can verify fix. Looks good to me!

Thanks for your contribution!

sqs:
This actually regressed something else for me:

If I am indented by 1 or more tabs and type any of `({[]}):`, my cursor jumps back 3-4 positions.

Undoing this commit fixes this issue.

josteink:
Cases like this are troublesome... We should try to create a TypeScript "reference-document" and setup some automated tests to ensure all known cases are still working when me make changes and avoid regressions.

(Like I've done with my own major-modes like [wsd-mode](https://github.com/josteink/wsd-mode/blob/master/wsd-tests.el#L176-L187) and [csharp-mode](https://github.com/josteink/csharp-mode/blob/master/csharp-mode-tests.el#L351-L362).

@ananthakumaran: I'd love to do this myself, but I'm quite pressed on time these days. Is this something you would be willing to look into? That way we would have a sure-fire way to test against any known regressions on any change. We could even setup a CI-build, so that these tests run on every single PR submitted.

hekt:
I'm sorry for that my commit causes the problem.

Following code seems resolve both problems. But this makes different behavior when run `(typescript-indent-line)` at beginning of line etc. I think it's better to revert my commit than this solution.

```elisp
(defun typescript-indent-line ()
  "Indent the current line as typescript."
  (interactive)
  (save-restriction
    (widen)
    (save-excursion
      (let* ((parse-status (syntax-ppss (point-at-bol))))
        (indent-line-to (typescript--proper-indentation parse-status))))))
```

ananthakumaran:
Sorry, I don't know much about the code base of typescript-mode and don't have much time to spend on it either. I only keep it on my account because it was not available in melpa before and tide mode has a dependency on it.

I would be happy if someone takes over the project. I myself can't spend much time other than occasional bug fixes and simple additions to the syntax highlighter.

ananthakumaran:
reverted this PR

PR #27, opened 2017-03-06
fix  'function*' indent in class

xcwen:
it's not good solution  ^_^
```
module.exports =function( app: App)  {
    return class _ extends ctrl_base( app) {
        * index() {
            const client1 = app.mysql.get("db1");
            const results= yield client1.query("show databases");

            return this.output_succ({"sdfall11" : 100});

        }
    }
};



```

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -1643,7 +1643,7 @@ See `font-lock-keywords'.")
   "Regexp matching keywords optionally followed by an opening brace.")
 
 (defconst typescript--indent-operator-re
-  (concat "[-+*/%<>=&^|?:.]\\([^-+*/]\\|$\\)\\|"
+  (concat "[-+/%<>=&^|?:.]\\([^-+/]\\|$\\)\\|"
           (typescript--regexp-opt-symbol '("in" "instanceof")))
   "Regexp matching operators that affect indentation of continued expressions.")
 


josteink:
Hey there. Thanks for the PR!

In theory they look good, and I would be happy to merge, but...

As a mere maintainer of this repo, I'm afraid of touching any indentation-code without a solid test-suite to back me up. That really goes for any Emacs-related indentation code and not just this project. :)

While it wouldn't be fair to push the burden of creating a test-suite on you, just to have this patch accepted, I think we should put in some effort to move in that direction too.

So a few questions:

* How long have you been using this patch? Have you observed any strange or unwanted side-effects since doing so?
* I honestly haven't used generators in TS yet, and I don't quite understand the provided code (extending callable classes?). Therefore I don't know what the absolutely "correct" form for this code should be. I can see how the current `typescript.el`-code indents it incorrectly though.
* I'm willing to create a reference-document and an associated ERT-test which attempts to verify the indentation of this document. ([Such as this](https://github.com/josteink/wsd-mode/blob/master/wsd-tests.el#L176-L187)). If I did that, would you be willing amend your patch to include a correctly formatted section which covers your use-case?

Also: @ananthakumaran: Could you get Travis CI setup for this repo too? No point in me creating tests if they don't get run :)

ananthakumaran:
enabled travis for this repo. Just add `.travis.yml` and it should work

josteink:
And there I've setup a CI-build which runs a basic reflow-test.

The test-document can probably be refined and cleaned up a little and what not, but for now it should be a lot better than nothing.

josteink:
@lddubeau: It was cases like this which I was certain would arise from this patch, but I wasn't 100% sure which ones it would be.

Thank you very much for helping out!

josteink:
Thanks for the patch, but this PR will be closed in favour of https://github.com/ananthakumaran/typescript.el/pull/29.

PR #28, opened 2017-03-07
Add CI and test-setup

josteink:
For now, the only tests we have are related to indentation, but that's still better than nothing :)


josteink:
No need for you to take any action, @ananthakumaran. I've just created this PR to verify that the CI-build actually runs.

And it runs beautifully. Very, very good. Consider this merged :)

PR #29, opened 2017-03-09
Fix indentation of methods that are generators.

lddubeau:
This is adapted from js.el's js--looking-at-operator-p function.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -1655,7 +1655,18 @@ See `font-lock-keywords'.")
          (or (not (looking-at ":"))
              (save-excursion
                (and (typescript--re-search-backward "[?:{]\\|\\_<case\\_>" nil t)
-                    (looking-at "?")))))))
+                    (looking-at "?"))))
+         (not (and
+               (looking-at "*")
+               ;; Generator method (possibly using computed property).
+               (looking-at (concat "\\* *\\(?:\\[\\|" typescript--name-re
+                                   " *(\\)"))
+               (save-excursion
+                 (typescript--backward-syntactic-ws)
+                 ;; We might misindent some expressions that would
+                 ;; return NaN anyway.  Shouldn't be a problem.
+                 (memq (char-before) '(?, ?} ?{ ?\;)))))))
+)
 
 
 (defun typescript--continued-expression-p ()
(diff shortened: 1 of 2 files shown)


typescript-mode.el
```diff
@@ -1655,7 +1655,18 @@ See `font-lock-keywords'.")
...
+               (looking-at "*")
+               ;; Generator method (possibly using computed property).
+               (looking-at (concat "\\* *\\(?:\\[\\|" typescript--name-re
+                                   " *(\\)"))
+               (save-excursion
+                 (typescript--backward-syntactic-ws)
+                 ;; We might misindent some expressions that would
+                 ;; return NaN anyway.  Shouldn't be a problem.
+                 (memq (char-before) '(?, ?} ?{ ?\;)))))))
+)
```
josteink: Could you explain the purpose of this section? It doesn't seem immediately obvious to me.

josteink:
Thanks for the PR and thanks for looking into this bug. I really hoped *someone other than me* would tackle it, and I'm glad you took one for the team :)

I also especially appreciate how you added all those test-cases. It makes it much easier for me to merge this with confidence, not to mention help prevent future regressions.

I think this looks very good, and apart from my one comment on the commit, I have no objections. Once we have that issue settled, I'll be happy to merge this.

typescript-mode.el
lddubeau: It ensures that we don't treat "*" as part of a generator expression unless it is preceded by the characters `(` `}` `{` or `;`. This is taken (comments and all) straight from `js.el` but I added `;`. The `;` is not needed for the latest incarnation of ES because classes only allow declaring methods. In TypeScript we can also declare fields of the class so we have one more case to account for.

This test is needed because the previous test that checks for the text that follows `point` is not completely determinative. It would yield a positive for instance, in this case when indenting the `foo()` line:

```
const q = 1 *
foo();
```

The case above cannot be handled properly without the last test. The comment indicates that there *are* cases where the final test will erroneously give a true result but that the syntax in those cases is so hairy as to not be worth the effort to get right. Here's an example:

```
const x = { }
*c();
```

As a matter of indentation the above is wrong because the asterisk cannot be anything else than a multiplication. Thus this is an expression continuation and the 2nd line should be indented. However, that expression, which would yield a `NaN` when computed, is not compiling in TypeScript due to type checking. 

Note that if you use `({} as any)` to get it to compile, the indentation is fine.

josteink:
Thanks for the clarification. Consider this merged!

PR #30, opened 2017-03-11
Don't interpret class members as keywords.

lddubeau:
Without this PR, having a method or field that is named "in" or "instanceof" or the same in a different case (e.g. "instanceOf") results in incorrect indentation.

For the record, I never name my methods "in", but IMO a method named "instanceOf" (with the capitalized "O") should be fair game. Without this PR "instanceOf" is treated as if it were the keyword "instanceof" and indentation is incorrect.

### DIFF
--- a/test-files/indentation-reference-document.ts
+++ b/test-files/indentation-reference-document.ts
@@ -161,5 +161,39 @@ namespace ts.server {
 
         const f = window
             .document;
+
+        const g = f
+            instanceof Object;
+
+        const h = "q"
+            in [1, 2];
+
+    }
+
+    {
+        // Object with fields that are keyword names.
+        const a = {
+            in: 1,
+            IN: 1,
+            instanceof: 1,
+            instanceOf: 1,
+        };
+
+        // Objects with methods that are keyword names.  At the top of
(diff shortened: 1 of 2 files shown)


josteink:
You're on a roll! Thanks for the patch and the associated test-case.

I've done some basic testing, I can reproduce the problem, and I can see that the patch fixes it (even though I honestly trust the CI to do a better job at that than I do).

My only objection, is that if you send in more patches in the future, and I see one more line of code which looks like this:

````elisp
 (memq (char-before) '(?, ?{ ?} ?\;)))))
````

Then, you'll have to abstract that away in a defun for clarity and reduced code-duplication. But that's for next time. ;)

Seriously though. Good work. Consider this merged!

lddubeau:
I did not refactor the code to have both the branch that deals with keywords and the branch that deals with generators share the same `memq` test because I was operating under the doubt that perhaps I had forgotten some scenario that would require the two to diverge, if not immediately upon inspection of the PR, then soon when someone points another case I did not think of. I meant to make a note of it in the PR but I forgot. :( 

Doing indentation without the benefit of an AST to know what's what is full of gotchas.

PR #35, opened 2017-04-24
Fix indentation of funcs with return annotations.

### LINKED ISSUE #8: Wrong indentation after multi-line function declaration with returned type
If a function declaration is broken on multiple lines AND has a returned type AND the opening brace is not on a line by itself, then there is a very wide indentation on the following lines.

``` js
function foo(firstParam: string,
             secondParam: number,
             thirdParam: boolean): boolean {
               console.log('look where this is indented');
               console.log('it is very strange');
             }
```

Expected:

``` js
function foo(firstParam: string,
             secondParam: number,
             thirdParam: boolean): boolean {
  console.log('look where this is indented');
  console.log('it is very strange');
}
```

lddubeau:
Without this patch, typescript-mode cannot handle functions that have a return type annotation.

Fixes #8.

### DIFF
--- a/test-files/indentation-reference-document.ts
+++ b/test-files/indentation-reference-document.ts
@@ -50,6 +50,8 @@ namespace ts.server {
         pid: number;
     }
 
+    export type SomeType = string | number;
+
     class Logger implements ts.server.Logger {
         private firstInGroup = true;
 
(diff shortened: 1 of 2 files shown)


josteink:
This looks like solid work, and the amount of test-cases you've put in is just staggering.

As for the code-changes themselves, there's quite a lot of new code you've been adding and I will need to set off some time to make sure I understand it properly.

I'm a bit short on time these days, so therefore I cannot make any promises as to when I can get this merged.

@ananthakumaran: Do you feel like landing this one? Seeing as this is mostly adding new code and not modifying existing one, I feel you may be better to judge if this is properly following conventions and stuff like that.

ananthakumaran:
I don't have much experience with the code base. I once did some cleanup, mostly related to repl code. But other than that, I don't have much idea. Please merge according to your will.

Judging from the amount of test case added, I am +1 on this patch.

lddubeau:
For all the test cases I have, I still managed to miss one case. I've just put in a fix and a new test case.

There's one thing that the patch does which may seem arbitrary. Previously, this indentation would be "correct" (by "correct" I mean that indenting all of the following lines with the mode would not change anything):

```
const a = [];
a.push({
    q: 1;
});
```

However, the above is now indented:

```
const a = [];
a.push({
           q: 1,
       });
```

This may look like an arbitrary change until one considers the following case. Before this patch:

```
function smurf(a: {}, b: {}) {}
smurf({
},
      {
      });
```

After the patch:

```
function smurf(a: {}, b: {}) {}
smurf({
      },
      {
      });
```

Before the patch, the 1st argument was treated inconsistently.

josteink:
I think a simple rule of what is deemed correct is what `tsserver` formats to by default (if you in `tide` set format-on-save to be true. That's what idiomatic typescript should look like, according to Microsoft. (Much like gofmt is convention for go).

We're not pixel perfect when it comes to compliance yet, but any improvements in that regard is surely a good thing.

I'll see when I can get time to look at this :)

josteink:
First of all, sorry about the hold up in reviewing this PR. I've been really busy :(

I've now looked over the code, and for most parts it looks good. I can't follow all the code everywhere, so I definitely appreciate the documentation provided for the functions.

Looking over the code and test-cases, the two things I was looking for (which for me has been a similar issue in `csharp-mode`) was the following:

* lambda expression (or arrow-functions)
* generics in parameter-lists, or more specifically *nested* generic types (`Array<FieldProcessor<int,string>>` etc).

From the provided test-cases, I can see lambda's are well covered.  However I only see use of generics in *return* values, and no use of nested generic types at all.

Do you think the provided code would be able to handle that properly? If so, could you add a test-case or two for that as well?

Other than that, I guess this looks fine and I'm happy to merge this patch.

And again: Sorry about the delay!

lddubeau:
No hard feelings here about the delay. I'm sure we have a lot of things demanding our time.

I just want to put in a quick comment: the logic I have here is not consistent with how `tsserver` (through `tide`) indents code. I intend to write a substantial comment about the issue when I have more time. (I thought I'd have time to write this substantial comment last weekend. Funny how that goes.) For now I'll just say that, upon initial inspection, I suspect that `tsserver` is doing the Right Thing(tm). @josteink I'll address your questions when I write the "substantial comment" I'm promising to write.

lddubeau:
Ok here's the promised comment... which is not going to be as substantial as I thought earlier because I changed my mind as I was researching the issue.

I found that my suspicion regarding `tsserver` was wrong. I don't like what it does. I found it to format code in a way that violates `tslint`'s `align` rule. I've been developing with this rule turned on and have not had issues caused by `typescript-mode.el` (or my patch).

This being said, I'm going to update the patch to eliminate the seemingly arbitrary difference in indentation introduced by my patch that I mentioned in [that earlier comment](https://github.com/ananthakumaran/typescript.el/pull/35#issuecomment-297794761). So the two cases shown in that comment will remain indented as they were before the PR.

josteink:
> This being said, I'm going to update the patch to eliminate the seemingly arbitrary difference in indentation introduced by my patch that I mentioned in that earlier comment. So the two cases shown in that comment will remain indented as they were before the PR.

That's all fine for me. Let me know when you're done and want someone to review and merge!

lddubeau:
Alright I've finally made the change I described in my earlier comment. I understand I can do a merge myself (thanks for that)... but does this  need review? (And once I'm ready to merge, should I *squash and merge* rather than just merge? The individual commits are probably not all that interesting to keep.)

josteink:
> once I'm ready to merge, should I squash and merge rather than just merge? The individual commits are probably not all that interesting to keep

That's your call. Me myself, I like atomic commits on the master branch, but I'm not always that good at sticking to it myself :)

> does this need review

For bigger changes, reviews are always good.

Sometimes I'll admit I don't understand the code, and will ask stupid questions. That will force you to explain it. And you can't explain something you don't fully understand.

Often I find being asked "stupid" questions by code-reviewers, forces me to think differently about the problem at hand, and that this alone is often enough to uncover some scenarios or use-cases I had originally not considered. This often leads me to realize I should probably change the code slightly.

In short: Don't be afraid to ask for reviews. I'll try to take a look at this one now :)

josteink:
Seems like a fairly well scoped change, with a impressive test-suite.

If I were to nitpick anything it would be that `typescript--backward-to-parameter-list` is a fairly large and involved function. Would it be possible to split it into a smaller functions? 

For instance, I see you have some complex tests deep inside the `loop`-macro, where the combination of these tests and nested conditionals makes getting the big picture of what the code is doing a bit hard to follow.

In theory splitting out these tests into separate functions could make them reusable for other parts of the codebase. Even if they eventually turn out not to do so, I think it would on overall help improve the semantics of the code anyhow.

Agreed? Disagreed?

lddubeau:
@josteink I looked at the code again just now. I'm going to add some comments to it because I think it would help with getting the big picture. On the other hand, I'm not seeing much benefit from splitting `typescript--backward-to-parameter-list` further. I had already parceled out `typescript--backward-over-generic-parameter-list` from it because I saw that one as being likely to be reused elsewhere, and the logic I have in there appears general enough to me for this function to be reused in other contexts. However, the rest of the logic in `typescript--backward-to-parameter-list` appears to me to be rather context-specific.

josteink:
Yeah. I guess this one has been nitpicked enough by now.

All in all I'm more than happy about it, compared to the regular PR on some of the projects I'm involved in :)

lddubeau:
I've rebased the fix to reduce it to two commits. I figured there may be value in keeping my flip-flopping about the indentation of function parameters as part of the commit history, in case it comes up again in the future.

PR #36, opened 2017-04-24
Support rule names and severity in tslint reports.

lddubeau:
This does two things:

1. Add support for the `verbose` formatter, which is the same as the default formatter but starts each warning with the rule name. For instance, where the default formatter would output
   ```
   src/modules/authenticator.ts[2, 26]: ' should be "
   ```
   The `verbose` formatter would have:
   ```
   (quotemarks) src/modules/authenticator.ts[2, 26]: ' should be "
   ```

2. Add support for the reports in tslint 5.x. The 5.x series distinguishes warnings from errors and uses the corresponding labels in its output.


josteink:
LGTM on first glance. I'll try to find time to get this reviewed and merged some time tomorrow.

josteink:
First of all: I really like how you've improved the overall code-quality by documenting it much better than it used to be, and also adding test-cases. I'm super-happy to merge patches like this.

On the somewhat less bright side, I did some basic testing... And I couldn't get anything to work. Not even stuff which used to work before.

After a while, and trying out running tslint from the command line (I couldn't get my output to match yours), I got a message saying I was running v4.0.1, and that I should/could update.

Reverting to the old code made the output from tslint 4 trigger just fine. Basically your improved support for tslint 5 makes us no longer support tslint 4, and that's not really too great either.

For the record, the output for a similar linting error looks like this in tslint 4:

````
/home/jostein/build/proj/src/file.ts[10, 25]: ' should be "
````

So that leaves us three choices as far as I see it:

* Keep the code as is and not support tslint 5.
* Apply your patch and break stuff for users still on tslint 4 without any explanation.
* Improve the patch to handle both output-formats.

With respect to keeping life simple for our users, I'm honestly only going to consider the last option a viable option right now.

I'll see if I can find some simple adjustments to your already improved regexp to make that happen.

lddubeau:
I must have goofed somewhere. The intent was always to support both tslint 4.x and 5.x at the same time. I'm on tslint 4.x myself (cannot upgrade until `tslint-microsoft-contrib` supports 5.x). Every single piece of code I submit for PR here is used in actual software development before being submitted. It definitely did work at some point but I may have screwed things up along the way and not realized it.

I'll take another look as soon as I can put time aside for it.

josteink:
The regexp looks fine to me and all the unit-tests pass so I'm going to assume that bit is in order. 

Maybe there's something wrong about how it gets wired up to compilation-mode?

lddubeau:
Found the problem, and put in a fix. The list of regex matches that I gave for compilation-mode to look at was wrong.

I could see if I can add a test that would catch what the problem was but it'll take me a while to come up with something sensible because I don't see a way to do this short of giving `compilation-mode` some mock data and see if it can make sense of it, and I'm not very familiar with how it works.

josteink:
Tested and works fine! Thanks for another quality patch!

Consider it merged :)

josteink:
As for how you can test this "properly"... I think it's reasonable to assume that the thing which will get meddled most with over time is the regexp itself. I therefore think having test-coverage for that is the most important thing.

I've done something similarly in `csharp-mode`: https://github.com/josteink/csharp-mode/blob/master/csharp-mode-tests.el#L137-L184

And as you can observe: It has the same short-coming as your test. It doesn't actually validate *how* things are added to `compilation-mode`.

It may not be perfect, but it's certainly better than nothing, and for now IMO more than good enough.

PR #46, opened 2017-08-13
Fix indentation for union type in return annotation

lddubeau:
This fixes a mistake in the indentation changes I made previously, and cleans up a leftover comment and the code a bit.


lddubeau:
I figure this is so trivial as to not need review.

josteink:
LGTM. As long as the test runs, I'm happy :)

Issue #50, opened 2017-08-25
Do we need the fontification hack?


lddubeau:
The mode initialization code contains this [hack](https://github.com/ananthakumaran/typescript.el/blob/813cf4fef30d95d055268c06afaf61b4b4f49878/typescript-mode.el#L2264-L2268):

```
  (let (font-lock-keywords)         ; leaves syntactic keywords intact
    ;; Avoid byte-compilation errors.  `font-lock-fontify-buffer' is
    ;; marked as interactive only in Emacs 25.
    (with-no-warnings
      (font-lock-fontify-buffer)))
```

The code above prevents users from having a `typescript-mode-hook` that immediately turns on `flyspell-prog-mode` and performs `flyspell-buffer` so that all spelling errors are immediately marked when `typescript-mode` starts. The problem with the code above is that it performs a fontification that only takes into account the syntactic classes of the text without applying the rules in `font-lock-keywords`. I've setup my flyspell predicate so that it requires the buffer to have been fontified at least up to the point it needs to examine. However, the first fontification triggered above does not contain the result of applying the rules in `font-lock-keywords`, which do matter for the predicate to work properly. So the initial `flyspell-buffer` run is running with incomplete fontification information, and ends up spell checking words that should not be spell-checked.

I worked around the problem by refontifying the buffer before starting `flyspell-prog-mode` and `flyspell-buffer`.

(Note: it would be *possible* to setup a predicate for flyspell that does not depend on fontification information. However, the predicate would get complicated *fast*. I'd rather avoid this if at all possible.)

I took a glance at `js-mode` and it looks like it *used* to perform an immediate fontification but that this immediate fontification was removed:

```
  ;; Important to fontify the whole buffer syntactically! If we don't,
  ;; then we might have regular expression literals that aren't marked
  ;; as strings, which will screw up parse-partial-sexp, scan-lists,
  ;; etc. and produce maddening "unbalanced parenthesis" errors.
  ;; When we attempt to find the error and scroll to the portion of
  ;; the buffer containing the problem, JIT-lock will apply the
  ;; correct syntax to the regular expression literal and the problem
  ;; will mysteriously disappear.
  ;; FIXME: We should instead do this fontification lazily by adding
  ;; calls to syntax-propertize wherever it's really needed.
  ;;(syntax-propertize (point-max))
```

[...]

PR #51, opened 2017-08-28
jsdoc fontification and spellcheck support

lddubeau:
This is the fleshed out version of 

https://github.com/ananthakumaran/typescript.el/tree/experiment/typedoc-tags

I've purposely split it into multiple commits:

1. There was an explicit invocation of the mode hooks in the code. I've removed it because `define-derived-mode` should call it automatically. (It sure does call it for me.) So the explicit invocation was redundant.

2. I've added fontification support for the jsdoc/typedoc tags. It is **very** unlikely to be the final word on this. The fact of the matter is that "jsdoc" is really a family of dialects for writing JavaScript documentation. Typedoc is one member of the family which focuses on the needs of TypeScript. Anyway, there's no definite set of tags for "jsdoc". Such and such version of jsdoc3 supports a definite set of tags, but some people use Google's Closure compiler which supports a different dialect, others run jsdoc2 (they really shouldn't but...) which supports another dialect, Typedoc supports yet another dialect, etc. I've named Elisp symbols so that those which deal with features usually common across dialects have `jsdoc` in their name, and those which are specific to Typedoc have `typedoc` in their name.

3. I've added a custom predicate for `flyspell-prog-mode`. In addition to the default fontification-based logic that `flyspell.el` provides, it excludes from spell-checking strings in `import` statements that refer to module names.

4. I've removed the fontification hack reported in #50.

I see there are CI failures on older versions of Emacs. I don't know yet what the solution for that is going to be.

I've used this code to spell check a fairly large code-base that uses Typedoc as its documentation processing tool.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2265,9 +2265,8 @@ Key bindings:
     ;; Avoid byte-compilation errors.  `font-lock-fontify-buffer' is
     ;; marked as interactive only in Emacs 25.
     (with-no-warnings
-      (font-lock-fontify-buffer)))
+      (font-lock-fontify-buffer))))
 
-  (run-mode-hooks 'typescript-mode-hook))
 
 ;;;###autoload
 (eval-after-load 'folding
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -414,6 +414,147 @@ Match group 1 is the name of the macro.")
    :paren-depth most-negative-fixnum
    :type 'toplevel))
 
+;; Note that all typedoc/jsdoc regexp by themselves would match occurrences that appear outside
+;; documentation comments. The logic that uses these regexps must guard against it.
+(defconst typescript-typedoc-link-tag-regexp
+  "\\[\\[.*?\\]\\]"
+  "Matches a typedoc link.")
+
+(defconst typescript-typedoc-literal-markup-regexp
+  "\\(`+\\).*?\\1"
+  "Matches a typedoc keyword markup.")
+
+;; This was taken from js2-mode.
+(defconst typescript-jsdoc-param-tag-regexp
+  (concat "\\(?:^\\s-*\\*+\\|/\\*\\*\\)\\s-*\\(@"
+          "\\(?:param\\|arg\\(?:ument\\)?\\|prop\\(?:erty\\)?\\)"
+          "\\)"
+          "\\s-*\\({[^}]+}\\)?"         ; optional type
+          "\\s-*\\[?\\([[:alnum:]_$\.]+\\)?\\]?"  ; name
+          "\\_>")
+  "Matches jsdoc tags with optional type and optional param name.")
+
+;; This was taken from js2-mode.
+(defconst typescript-jsdoc-typed-tag-regexp
+  (concat "^\\s-*\\*+\\s-*\\(@\\(?:"
+          (regexp-opt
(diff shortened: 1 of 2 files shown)


josteink:
Looks like extensive and proper work, as usual. :)

I fully support these changes.

lddubeau:
I've added commits to fix the issues we discussed. I've also run spell check on another code base, smaller than the earlier one I used, but I still managed to find a couple issues that I fixed.

One change I forgot to mention in my initial PR: the fontification includes a regexp for tslint flags:

```
// tslint:disable-next-line:blahblah
```

from the word `tslint` onwards this is fontified using `font-lock-preprocessor-face` because to me it is akin to a preprocessor directive... though maybe there are good reasons to want another face for it. Let me know if you have an issue with this.

Once we're in agreement, I plan to rebase the PR here to fold the additional commits I just added into the original 4 commits and merge.

And I just got notice that the tests passed. :tada:

typescript-mode.el
```diff
@@ -1741,8 +1741,11 @@ and searches for the next token to be highlighted."
...
 (defun typescript--flyspell-mode-predicate ()
   "A custom predicate to help `flyspell-prog-mode' determine whether a word should be checked."
-  ;; We depend on fontification for our results.
-  (font-lock-ensure)
+  ;; We depend on fontification for our results. font-lock-ensure is defined on
+  ;; Emacs 25 and over. Earlier versions use font-lock-fontify-buffer.
+  (if (fboundp 'font-lock-ensure)
+      (font-lock-ensure)
+    (font-lock-fontify-buffer))
   (and
```
josteink: This is nice. I like this way better.

It's even documented so at whatever point in the future when we don't feel the need to support Emacs 24.x we can clean it up.

typescript-mode.el
```diff
@@ -1756,12 +1756,12 @@ and searches for the next token to be highlighted."
...
-          (and string-start-pos
-               (save-match-data
-                 (save-excursion
+   (save-excursion
+     (not (let* ((parse (syntax-ppss (1- (point))))
+                 (string-start-pos (and (nth 3 parse)
+                                        (nth 8 parse))))
+            (and string-start-pos
+                 (save-match-data
                    ;; Move to back to the start of the string, then past any ws
```
josteink: This is absolutely nitpicking, and I in no way mean to imply the tests so far are insufficient.

That said: When you encountered this issue and decided to write a fix for it... Was this a regression that can somehow be verified in a test, and thus prevented in the future?

josteink:
Pristine as always. 

If you can fix up and clean the commits like you said, this is about as perfect as I can see a PR being.

Go do your thing :)

typescript-mode.el
lddubeau: I noticed the issue when I was doing `M-: (typescript--flyspell-mode-predicate)` to figure out one of the issues I ran into. I noticed that the cursor was moving back, and it was annoying me.

I did not experience a failure while using flyspell through the normal means.

I guess I could still ensure in the test suite that point is not moved by the predicate, which is probably a sensible rule, generally speaking.

typescript-mode.el
lddubeau: I had a vision of a future me some years from now looking at this code and going "Why the hell did I do this?" So I commented. :)

Issue #53, opened 2017-10-07
Highlighting for exported functions.


jack-williams:
This isn't a bug so much as a request / query. I'm just wondering about syntax highlighting for functions that are exported.

Given the following example:

```
function foo() {

}

export function bar() {

}
```

`foo` will be coloured but `bar` wont. Was the a conscious choice? Personally I'd like both function names to be coloured (probably using the same colour). I have a hacky solution that changes `typescript--function-heading-1-re` to include an optional `export` string in the regex, however I'm really not knowledgeable enough to know if this is the wrong approach.

(And thanks for the great work on the mode!)

josteink:
Thanks for the bug-report.

I'm going to assume this was an omission. I'll label it a bug, and see if we can't get it fixed.

josteink:
Thanks for the bug-report.

I'm going to assume this was an omission. I'll label it a bug, and see if we can't get it fixed.

If you could show us your changes, maybe we could review them together and see if we find them sufficient. If so, you could always submit them as a PR.

jack-williams:
I've added a new matcher for declare or export modifiers:

```
(defconst typescript--module-declaration-re
  "^\\s-*\\(?:declare\\|\\(?:export\\(?:\\s-+default\\)?\\)\\)?"
  "Regexp matching ambient declaration modifier or export declaration")
```
and then added that to `typescript--function-heading-1-re` as follows:

```
(defconst typescript--function-heading-1-re
  (concat
   typescript--module-declaration-re
   "\\s-*function\\s-+\\(" typescript--name-re "\\)")
  "Regexp matching the start of a typescript function header.
Match group 1 is the name of the function.")
```

josteink:
Sorry about the delay. I've had quite a few other issues I've been trying to close before starting yet another one. :)

Comparing to what is already done in `js-mode`, you have the same exact problem there, so there is no code we can "borrow" for a more proven approach.

`js2-mode` seems structured in a completely different way, and while it has this already solved, I see no obvious way to re-use their solution.

As such, and from what I can tell, that looks perfectly OK. I have no objections to it.

Feel free to send it in as a PR, and other people more versed in font-locking code can help improve it if they see the need.

jack-williams:
Great, I'll put a PR in over the next few days. Thanks!

josteink:
Fixed in #54. Closing.

PR #54, opened 2017-10-11
Highlight function name for exported or declared functions.

jack-williams:
Addresses the following issue (#53):

Extends syntax highlighting for function names to the following cases:

```
export function foo() { }
export default function foo() { }
declare function foo();
```
specifically: when `export`, `export default`, or `declare` precedes a function definition.

### DIFF
--- a/typescript-mode-tests.el
+++ b/typescript-mode-tests.el
@@ -192,11 +192,16 @@ as the search pattern."
 (defun font-lock-test (contents expected)
   "Perform a test on our template. `CONTENTS' is the string to
 put in the temporary buffer. `EXPECTED' is the expected
-results. It should be a list of (LOCATION . FACE) pairs."
+results. It should be a list of (LOCATION . FACE) pairs, where
+LOCATION can be either a single location, or list of locations,
+that are all expected to have the same face."
   (test-with-temp-buffer
    contents
    (dolist (spec expected)
-     (should (eq (get-face-at (car spec)) (cdr spec))))))
+     (if (listp (car spec))
+         (dolist (source (car spec))
+           (should (eq (get-face-at source) (cdr spec))))
+       (should (eq (get-face-at (car spec)) (cdr spec)))))))
 
 (ert-deftest font-lock/documentation-in-documentation-comments ()
   "Documentation in documentation comments should be fontified as
@@ -256,13 +261,6 @@ new line after the start of '/**'."
      ("@param" . typescript-jsdoc-tag)
      ("meow" . typescript-jsdoc-value))))
 
-(defun copy-test-n (n location-root expected)
-  (let (tests)
-    (dotimes (i n tests)
-      (setq tests
-            (cons (cons (concat location-root (number-to-string i)) expected)
-                  tests)))))
-
 (ert-deftest font-lock/function-definition-prefixes ()
   "Tests that function names are highlighted in definitions, even
 when prefixed with module modifiers."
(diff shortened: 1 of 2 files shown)


typescript-mode-tests.el
```diff
@@ -256,18 +256,32 @@ new line after the start of '/**'."
...
+    '(("basicDefn" . font-lock-function-name-face)
+      ("exportedDefn" . font-lock-function-name-face)
+      ("exportedDefaultDefn" . font-lock-function-name-face)
+      ("declareFunctionDefn" . font-lock-function-name-face))
+    (copy-test-n 4 "x" font-lock-variable-name-face)
+    (copy-test-n 4 "y" font-lock-variable-name-face)
+    (copy-test-n 4 "xty" font-lock-variable-name-face)
+    (copy-test-n 4 "yty" font-lock-variable-name-face)
+    ;; Return types are not highlighted currently.
+    (copy-test-n 4 "ret" nil))))
```
josteink: Looking over this particular section of code (`copy-test-n`), I had to read it a few times for the intent to be clear, and for me to be sure I understood what it represented.

I prefer to have tests as self-explanatory as possible, so if (or when) a test fails, it should be immediately be what it failed and why it failed.

I see that there's some repetition in code-form in the test-data... And that you're trying to minimize the amount of code-duplication this causes in your test-code. Which is all fine.

But would it be possible to make some changes to this code to make the intent more clear?

* A better function-name for copy-test-n
* Create some more descriptively named abstractions on top of copy-test-n
* Restructure how you do the test repetitions?

How about something like:

````elisp
(create-test-case 'font-lock-variable-name-face "x0" "x1" "x2" "x3")
(create-test-case 'font-lock-variable-name-face "y0" "y1" "y2" "y3")
(create-test-case 'font-lock-variable-name-face "xty0" "xty1" "xty2" "xty3")
(create-test-case 'font-lock-variable-name-face "yty0" "yty1" "yty2" "yty3")
(create-test-case nil "ret0" "ret1" "ret2" "ret3")
````

Or even less verbose (if you feel this is too "heavy"):

````elisp
(create-test-case 'font-lock-variable-name-face
                  "x0" "x1" "x2" "x3"
                  "y0" "y1" "y2" "y3"
                  "xty0" "xty1" "xty2" "xty3"
                  "yty0" "yty1" "yty2" "yty3")
(create-test-case nil
                  "ret0" "ret1" "ret2" "ret3")
````

To me, that would make the intent lighting-clear. It would come at cost of some variable-name repetition, but I honestly think that would make it a better test.

File: typescript-mode-tests.el [hunk 2]
```diff
@@ -256,6 +256,33 @@ new line after the start of '/**'."
      ("@param" . typescript-jsdoc-tag)
      ("meow" . typescript-jsdoc-value))))
 
+(defun copy-test-n (n location-root expected)
```
lddubeau: This function would benefit from a docstring. I'd also suggest changing the name to something more descriptive like `make-test-variations`. Until I read the code, I thought it was just producing multiple copies of the same test, like a photocopier makes copies of a sheet.

(I see josteink made a similar comment.)

typescript-mode-tests.el
lddubeau: The work done by `copy-test-n` was specialized enough that it made sense to have a separate function. However, the code snippets offered in suggestion are fine, but we're going to do that, we might as well just modify `font-lock-test` to accept `(LOCATION . FACE)` with `LOCATION` being a list of string or numbers in addition to the basic case where `LOCATION` is a string or number. Then there's no need to have a separate function.

jack-williams: Thanks for the comments - I agree. I might have a go at changing `font-lock-test` to also accept `LOCATION` as a list. If this doesn't work out I'll just run with the suggestions above.

josteink:
Hey Jack.

Any news on this one? :)

jack-williams:
Sorry I've been away. I have changed `font-lock-test` to also accept a list of locations and updated my test case accordingly.

josteink:
No worries. I was just checking. Thanks for responding swiftly!

Anyway, I like this new revision a lot better. The tests look much cleaner now. I'm happy to merge this, unless anyone else have any other objections.

josteink:
No objections? Consider this merged then!

Thanks for the patch!

Issue #44, opened 2017-07-20
font highlighter ends regexp prematurely


lddubeau:
Given the following regexp:

```
/[a/b]/
```

The regexp will be highlighted as a regexp from the first forward slash, until the 2nd forward slash. However, the regexp really ends at the last forward slash. (For the benefit of readers who may not know: a forward slash appearing in `[...]` is treated as *part of* the regexp instead of treated as *terminating* the regexp. The regex above matches the characters `a`, `/` and `b`.)

I guess it is the twin of #34.

This issue does not annoy me enough to get to work on a solution in the short term so anyone is welcome to fix it and reap the glory. :)

josteink:
I've tested and this also seems fixed in PR #45.

josteink:
With PR 45 closed and not merged, I guess we will need to seek an alternate solution to this issue.

nnabinh:
Up, pls fix this

josteink:
PRs are welcome :)

Issue #34, opened 2017-04-21
String replace regex makes all the following lines as a string


hexa00:
Something like:

test.replace(/\\/g,"/");
nextlines are as if part of a string

Using : typescript-mode-20170324.1301

lddubeau:
The mode is working correctly. It is your regex that is wrong and throwing you off. 

You expect the first `/` to terminate your regular expression, but you have `\` before it, which escapes its normal meaning. You think your regular expression is:

```
/\/g
```

but it is:

```
/\/g,"/
```

and then it is followed by the start of a string which you never terminate.

The regex you want appears to be this: `/\\/g`. This will escape the escape character and make it match a backslash.

hexa00:
oops seems like the github comment missed a \ here... I should have enclosed it

The real code is like you said:
`test.replace(/\\/g,"/");`
With the same result...

lddubeau:
Ah, yes, I can reproduce it here now.

hexa00:
:) Note I tested in the JavaScript major mode and it works there...

josteink:
I did some basic testing and landed on this:

````
test.replace(/a/g,"/");    // good
test.replace(/a\a/g,"/");  // good
test.replace(/a\\a/g,"/"); // good
test.replace(/a\\/g,"/");  // not good
````

Basically if we have a `\` of any sort preceede the finalizing `/`, our code breaks. Not having looked at the code in question, it seems like some naive escape-handling code fails to realize detect the double `\\` as one escaped character, and thinks we're trying to escape the `/`.

**Comment by sudarshang** [comment]
Looking into how js-mode handles this

https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/js.el#L1725
```elisp
(defun js-syntax-propertize-regexp (end)
  (let ((ppss (syntax-ppss)))
    (when (eq (nth 3 ppss) ?/)
      ;; A /.../ regexp.
      (goto-char (nth 8 ppss))
      (when (looking-at js--syntax-propertize-regexp-regexp)
        ;; Don't touch text after END.
        (when (> end (match-end 1))
          (setq end (match-end 1)))
        (put-text-property (match-beginning 1) end
                           'syntax-table (string-to-syntax "\"/"))
        (goto-char end)))))
```

It seems to use a function to deal with regexes instead of using the regular expression that typescript mode uses

https://github.com/ananthakumaran/typescript.el/blob/master/typescript-mode.el#L1624
```elisp
[...]

PR #55, opened 2017-10-17
Simplify regexp literal

bkruczyk:


josteink:
There's always weird edge-cases when parsing is involved. 

That said, the code in question here has been unchanged since the initial commit, with none of the current maintainers altering it. 

For me at least, that means I don’t know any specific reasons to keep it as is, in its current form. 

I’d be happy to accept this PR if you can:

1. Prove it fixes a bug and
2. Don’t break anything else. 

And the best way to do that would be adding some test-cases to the test-suite for regexp-parsing. 

Take a look at the existing tests and let us know if you need any help.

bkruczyk:
@josteink thanks for feedback, maybe we can discuss test cases when I will update the PR, need a day or two

josteink:
Sure thing. Take your time.

bkruczyk:
If anyone have anything to add on added tests, you're welcome to leave a comment :)

josteink:
Thanks for the update!

When it comes to unit-tests I usually check one thing before considering myself "happy": I ensure that the test prove the error-condition by *failing* before the appropriate fix is applied.

In this case your test are equally good before and after your fix is applied, meaning they are not sufficient to reproduce the error you say you want fixed.

If you could you address that (that is, make the tests fail without your fix in place), I'll consider this good and will be happy to merge!

bkruczyk:
That's a great tip on testing "fixes", I'll incorporate it in my daily work, thanks!
I've updated the PR.

josteink:
Now that's more like it! Now the tests behave as tests should :)

And given my promise just one comment away, sorry about being a bit thick-headed here today (got a cold and all), but had the impression that this PR was meant to fix either #34 or #44.

As far as I can see from my interactive testing that's not the case though. Or is there something else I have misunderstood? Was this PR mostly about simplifying existing code?

If I try the testcase from #34:

````typescript
test.replace(/\/g,"/");
this_identified_is_still_a_string
````

Second line is still highlighted as a string somehow. Is there something wrong with my test-setup perhaps?

Again: Sorry for being thick-headed about all this when you're obviously putting in a good effort just trying to full-fill my requests.

bkruczyk:
Yup I intended for this PR to be a fix to #34. 

I think the highlighting is correct, please check [this comment](https://github.com/ananthakumaran/typescript.el/issues/34#issuecomment-296190355) :)

josteink:
Ah gotcha. My bad.

Thanks for your contribution and consider this merged!

Issue #40, opened 2017-06-08
Doesn't indent multiple variable declarations


raxod502:
`typescript.el` indents like this:

```
var x = 5,
y = 6;
```

I posit that it should indent like this:

```
var x = 5,
    y = 6;
```

josteink:
> I posit that it should indent like this

As far as Typescript goes, there's just *one* canonical answer to how things should be: However `tsserver` formats it.

And lucky for you, tsserver agrees :)

The good news/workaround: If you use typescript-mode with [tide](https://github.com/ananthakumaran/tide), it will use tsserver to reformat the document on each save, giving you the correct formatting.

It would of course be nice if we could make typescript-mode also indent like this out of the box. If anyone feels like making a PR which fixes this, I'll be happy to review.

dgillis:
Have at look at that pull request. It does not handle the case of multiline destructuring assignments. E.g., I think they ought to be indented like this:

```javascript
var a = 1,
      {x,
       y,
       z} = object;
```
But typescript mode is still aligning the continuations with the opening brace:
```javascript
var a = 1,
      {x,
      y,
      z} = object;
```
I have been using advice that fixes both cases. However there are a lot more edge cases to destructuring that I'm probably ignoring so I excluded.

I notice that typescript mode does correctly indent multiline desctructuring statements when they are on their own. So it would probably not be too difficult to generalize that in a way that it just gives the correct offset relative to the opening of the destructure (which could then be added to the normal multiline to get the proper indentation).

rhz:
What is the state of this issue? It would be really cool if typescript-mode would indent multiple variable declarations as in the opening comment, even if it doesn't get destructuring statements right in all edge cases.

lddubeau:
@rhz PR #56 was the only attempt at fixing the issue, and it died on the vine. Whoever wants to see this resolved should take a look at that PR and the comments on it and build from that to submit a new PR.

rhz:
@lddubeau I do not have the skills required nor any familiarity with the codebase to be able to do that. I just wanted to open an issue with a feature request and realised it existed already, so I popped in to see if there were still plans of adding this feature.

Issue #41, opened 2017-06-08
Support for jsdoc-like comments


raxod502:
As far as I know, TypeScript tends to be documented like this:

```
/**
 * Run the provided promise, and call process.exit with an appropriate code.
 */
export function runPromise<T>(promise: Promise<T>): void {
```

In `js2-mode`, I can type:

```
/**
```

and press `C-M-j`, to get:

```
/**
 * 
 */
```

with point at the end of the second line. This would be a nice feature to have in `typescript.el`, as well.

josteink:
From the docs:

> C-M-j runs the command indent-new-comment-line (found in global-map), which is an alias for ‘comment-indent-new-line’ in ‘newcomment.el’.

Running this command anywhere in typescript-mode gives mixed results:

````

var x;
|<- newline here

//
//|<- newline here

/*
/*|<- newline here

/**
/*|<- newline here
````

If we're going to override this command with a custom command for typescript-mode, we should properly address all these comment-styles.

ananthakumaran:
I wonder whether we should rebind `M-j` to `c-indent-new-comment-line` by default. It seems to work better than `indent-new-comment-line` from my minimal testing.

josteink:
That's a good idea, and leaves us reinventing fewer wheels than we need to.

Have you experienced any cases where that does anything which is explicitly *wrong*?

raxod502:
It might also be relevant to ask @dgutov if there is a reason that `js2-mode` doesn't do this, and instead implements its own `js2-line-break` function.

ananthakumaran:
> Have you experienced any cases where that does anything which is explicitly wrong?
It's not as good as js2-mode, but better than what we have. I never seen encountered any case so far where indent-new-comment-line is better

dgutov:
In js2-mode, it's wasn't a particular decision to write `js2-line-break` instead of using `c-indent-new-comment-line`.

Before I got to it, `js2-line-break` was a command `js2-enter-key` bound to `RET`. So using a custom `'comment-line-break-function` function provides some continuity.

So `c-indent-new-comment-line` might be a good choice of you, except it's tying you to CC Mode a bit, and will probably require you to set a few related variables.

liyaodong:
FYI `tide-jsdoc-template`can insert now

PR #56, opened 2017-11-03
Fix multiline variable declarations.

dgillis:
See #40

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2120,6 +2120,26 @@ moved on success."
     (when location
       (goto-char location))))
 
+(defun typescript--multiline-declaration-indentation ()
+  "Returns the proper indentation for lines which are continuations of
+multiline var/let/const declarations. If the point is not on such a line,
+returns nil."
+  (save-excursion
+    (beginning-of-line)
+    (let ((decl-re "^ *\\(var\\|const\\|let\\)\\_> *")
+          (is-looking-at-decl-keyword nil))
+      (save-match-data
+        (while (and (looking-back ",[[:space:]\n]+" nil)
+                    (= (forward-line -1) 0)
+                    (not (setq is-looking-at-decl-keyword (looking-at decl-re)))))
+        (when is-looking-at-decl-keyword
+          ;; Found the var/let/const declaration of which the initial line was a
+          ;; continuation of. The correct indentation is therefore the relative
+          ;; offset of the word following the var keyword.
+          (let ((bol-pos (point)))
+            (re-search-forward decl-re)
+            (- (point) bol-pos)))))))
+
 (defun typescript--proper-indentation (parse-status)
   "Return the proper indentation for the current line."
   (save-excursion
@@ -2130,6 +2150,7 @@ moved on success."
           ((typescript--ctrl-statement-indentation))
           ((eq (char-after) ?#) 0)
           ((save-excursion (typescript--beginning-of-macro)) 4)
+          ((typescript--multiline-declaration-indentation))
           ((nth 1 parse-status)
            (let ((same-indent-p (looking-at
                                  "[]})]\\|\\_<case\\_>\\|\\_<default\\_>"))


josteink:
Right now the CI build fails due to the following test failing: `indentation-reference-document-is-reflowed-correctly`.

Basically if you're updating the indentation-rules, you should:

1. update the test reference-document accordingly, where you think it should be affected.
2. verify that no other formatting-rules have been broken in the process. (`make test`)

Please check what's the cause of the CI-failure, and add a test-case for the indentation-rule you're implementing if not already there. If there's such a rule already, please update it to stop the old tests from crashing the build :)

josteink:

I'll be happy to look at it, but first we should probably get the CI build sorted out.

Right now the CI build fails due to the following test failing: `indentation-reference-document-is-reflowed-correctly`.

Basically if you're updating the indentation-rules, you should:

1. update the test reference-document accordingly, where you think it should be affected.
2. verify that no other formatting-rules have been broken in the process. (`make test`)

Please check what's the cause of the CI-failure, and add a test-case for the indentation-rule you're implementing if not already there. If there's such a rule already, please update it to stop the old tests from crashing the build :)

dgillis:
It is failing because assignment expressions where the right-hand side value is spread over multiple lines is being mistakenly interpreted as a new assignment on the second (and onward) lines. E.g., `b` ought to be aligned with `a` below but instead my function is doing this:

```js
const foo9 = (a: string,
      b: string): Array<number> => {
    return [1];
}
```
This is a similar issue to indenting multiline destructuring assignments (see my comment on #41). In both cases, I think the rigourous way to do it is to have function that can walk skip back over expressions of the form `<LHS> = <RHS>,` (or simply <LHS>, when it is just a name declaration) until it finds the var/let/const declaration. The beginning of each subsequent <LHS> should then be aligned to the first one after the var/let/const.

Lines on the interior of multiline LHS/RHS expressions are more complex. However, typescript-mode otherwise handles them correctly. So I can imagine there is some way to utilize that functionality and augment it with an offset based on the indentation of the parent LHS/RHS expression. I do not have much familiarity with the codebase so it's hard for me to come up with a way of doing this that does not seem like a lot of reinventing the wheel.

josteink:
Sorry about the delay. I've been quite busy.

And seeing some things being broken, I wanted to take a proper look at things to ensure I had  the full picture before I gave some feedback (naturally postponing things even more...)

Well, now I've taken a *brief* look at least.

So like you've correctly assessed, the CI-build fails because indentation for functions with return annotations no longer behaves as they used to.

This was implemented and fixed by @lddubeau [quite recently](https://github.com/ananthakumaran/typescript.el/pull/46), so I don't think it will be too popular if we break that while attempting to fix something else.

Do you think it would be possible for you to implement your fix *without* causing the existing behaviour to break?

> I do not have much familiarity with the codebase so it's hard for me to come up with a way of doing this that does not seem like a lot of reinventing the wheel.

Honestly most work I do here is as a maintainer, more than actively writing code, so I can definitely sympathize.

@lddubeau : Would you be able to give this leads to follow? Have you implemented any such functions like the one he asks about?

lddubeau:
It seems to me that the failing case that was quoted above is failing because the new code is looking at the test case and treating it just like a multiline declaration, which it isn't. And indeed, just looking at the code, I see the new functionality runs before any of the code I added to handle type annotations for return value. This could be assessed by stepping through the old code and through the new code while trying to manually indent the faulty line in the test file. (I use `edebug` for this. I do `C-u C-M-x` to recompile with `edebug` the function I want to step through, and I recompile it with `C-M-x` when I'm done. However, I would not be surprised if there's a nicer way to do it.)

It may be possible to keep the call to `typescript--multiline-declaration-indentation` in the same location it is in right now. However, `typescript--multiline-declaration-indentation` would have to be modified to be more restrictive as to what it consider to be a multiline declaration.

However, if I were coding this, I would move that call down into the `(nth 1 parse-status)` branch. As a general rule, there's very very very very very little that can easily go before the `(nth 1 parse-status)` branch of the `cond` in `typescript--proper-indentation`. The initial tests slice the pie in very large and mostly unambiguous pieces: am I in a string?, am I in a comment?, is this a cpp macro (really?... barf!)?, etc. The exception is the test with `typescript--ctrl-statement-indentation` which is more complex, but, well, it works. I'd expect the new code to operate from within the code handling `(nth 1 parse-status)`, perhaps in a way analogous to how that branch uses `same-indent-p` and `continue-expr-p`.

----

Question unrelated to the above. Can the new code handle this?

```
var x = () => {
    var foo = 1,
        moo = 2;
},
    y = () => { };
```

The above is formatted according to `tsserver`.

typescript-mode.el
```diff
@@ -2120,6 +2120,26 @@ moved on success."
...
+          (is-looking-at-decl-keyword nil))
+      (save-match-data
+        (while (and (looking-back ",[[:space:]\n]+" nil)
+                    (= (forward-line -1) 0)
+                    (not (setq is-looking-at-decl-keyword (looking-at decl-re)))))
+        (when is-looking-at-decl-keyword
+          ;; Found the var/let/const declaration of which the initial line was a
+          ;; continuation of. The correct indentation is therefore the relative
+          ;; offset of the word following the var keyword.
+          (let ((bol-pos (point)))
```
lddubeau: `bol` to me means "beginning of line". Sometimes, this point *is* the beginning of line, but not *always*, right? If that's the case, I'd give it a different name. `keyword-start` maybe??

typescript-mode.el
```diff
@@ -2120,6 +2120,26 @@ moved on success."
     (when location
       (goto-char location))))
 
+(defun typescript--multiline-declaration-indentation ()
+  "Returns the proper indentation for lines which are continuations of
+multiline var/let/const declarations. If the point is not on such a line,
+returns nil."
+  (save-excursion
+    (beginning-of-line)
+    (let ((decl-re "^ *\\(var\\|const\\|let\\)\\_> *")
```
lddubeau: I see another regex searching for `var`, `const`, `let` elsewhere in `typescript--font-lock-keywords-3`. Can we perhaps combine the two regexes into  a single `defconst` and refer to the new constant, so as to avoid duplication?

typescript-mode.el
```diff
@@ -2120,6 +2120,26 @@ moved on success."
...
+      (save-match-data
+        (while (and (looking-back ",[[:space:]\n]+" nil)
+                    (= (forward-line -1) 0)
+                    (not (setq is-looking-at-decl-keyword (looking-at decl-re)))))
+        (when is-looking-at-decl-keyword
+          ;; Found the var/let/const declaration of which the initial line was a
+          ;; continuation of. The correct indentation is therefore the relative
+          ;; offset of the word following the var keyword.
+          (let ((bol-pos (point)))
+            (re-search-forward decl-re)
```
lddubeau: If `is-looking-at-decl-keyword` is true, then `match-end` should already contain the point you are looking for. Is there something I'm missing? If I'm right then the whole `when` block could be reduced to:

```
(when is-looking-at-decl-keyword
  (- (match-end) (point)))
```

or something of that order.

typescript-mode.el
dgillis: Good point.

josteink:
Any updates on this PR? Any plans on fixing the things it breaks? :)

lddubeau:
I'm closing this due to lack of activity. I'll gladly reopen if development of the PR resumes.

PR #58, opened 2017-11-08
Fix/regex regex

### LINKED ISSUE #57: commented lines are sometimes treated as non-comments
This causes errors with both font-lock and indentation.

This can be demonstated in the [indentation reference document](https://github.com/ananthakumaran/typescript.el/blob/30391a518a1d41d5c534a30850a50b0974132aec/test-files/indentation-reference-document.ts#L34-L35) where the following lines will be syntax-highlighted (and have syntax-properties) as if they were not commented:

````typescript
        //     case "linux":
        //     case "android": {
````

I can reproduce this on Emacs 25.3 (Fedora) and Emacs git master.

When trying to pinpoint where it goes wrong using `font-lock-studio` it seems none of the expressions picked up by `font-lock-studio` contributes to this error-condition, but instead it is immediately reproduced after exiting it (but then only the line with `"linux"` in it).

Could this be related to something we explicitly override in our mode outside the regular font-lock and syntax constructs, like `syntax-begin-function`?

**Edit:** Can also reproduce on Emacs 25.1 on Windows.

### PR DESCRIPTION
**Pull request by lddubeau**
Fixes #57

### DIFF
--- a/typescript-mode-tests.el
+++ b/typescript-mode-tests.el
@@ -284,22 +284,33 @@ declare function declareFunctionDefn(x3: xty3, y3: yty3): ret3;"
                     '(("=" . nil) ("/foo/" . font-lock-string-face)
                       ("(" . nil) ("/bar/" . font-lock-string-face)
                       ("," . nil) ("/baz/" . font-lock-string-face)
-                      (":" . nil) ("/buzz/" . font-lock-string-face)))))
+                      (":" . nil) ("/buzz/" . font-lock-string-face))))
+  ;; Make sure that escaped forward slashes are handled too.
+  (font-lock-test "var a = /flip\\/flop/;"
+                  '(("=" . nil)
+                    (("/flip" "\\\\" "/" "flop/") . font-lock-string-face)
+                    (";" . nil)))
+  ;; A sequence of two forward slashes is never a regex, so there is
+  ;; no such thing as an \"empty regex\" when we use the forward slash
+  ;; notation.
+  (font-lock-test "=//g something // comment"
+                  '(("g something" . font-lock-comment-face))))
 
 (ert-deftest font-lock/text-after-trailing-regexp-delim-should-not-be-fontified ()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by josteink** [comment]
Wow. That was fast. Thanks!

While your approach clearly works... And we now have a test-case to prove it (which is *good*!)... Do you see any cases which this solves which wouldn't be solved by the simpler diff?

````diff
-  "[=(,:]\\(?:\\s-\\|\n\\)*\\(/\\)[^/]*\\(/\\)"
+  "[=(,:]\\(?:\\s-\\|\n\\)*\\(/\\)[^/]+\\(/\\)"
````

I'd like to keep things as simple as possible, and I have to admit that I'm having a little trouble following the changes you've made (Emacs regexp being as reader-friendly as they are and all that stuff).

That simpler diff also passes your test, so unless there's some other weird edge-cases I know about, that should be equally good, right?

**Comment by lddubeau** [comment]
The change you propose does not handle escaping forward slashes. For instance, in this example:

```
var a = /flip\\/flop/;
```

everything after the regular expression would get the string face. So the `;` and everything after it, if there more more code on the same line, would appear as a string.

The changes I made were to make the pattern disallow `//` as a regular expression (so require at least one character between the two forward slashes) and reintroduce handling of `\`. Basically, `\` followed by anything is considered part of the regex. This is *especially* important when you have the sequence `\/`, which means that `/` is not the regex terminator.

Your comment made me realize we need a test specifically for that.  So I've added one just now. I plan to squash everything into a single commit at merge time.

----

This was a fast submission because I opened a file where the bug made a total mess of things. The file was basically uneditable with Emacs. I had to put out that fire before I could resume work.

josteink:
> The change you propose does not handle escaping forward slashes.

Ah gotcha. You're indeed right about that. Also  your new test-case does indeed prove that my simpler regexp is insufficient.

I guess that means I'm 100% happy about these changes, as is.

Unless anyone else finds anything objectionable, I'm all for merging these changes.

PR #60, opened 2017-12-11
Fix handling of array shorthand messing up indentation, and simplify typescript--backward-to-parameter-list

lddubeau:
The first commit fixes indentation when there is an array shorthand (`foo[]` by opposition to `Array<foo>`) in a return value type.

The 2nd commit is a simplification of `typescript--backward-to-parameter-list`. As I was producing the fix in the first commit, I figured that there ought to be a way to simplify the code. If I were just submitting the 1st commit, I'd just push it to `master`, but I want to give interested people an opportunity to comment on the simplification (2nd commit), if desired.

If I don't hear anything 48 hours from this post, I'll merge into `master`.

### DIFF
--- a/test-files/indentation-reference-document.ts
+++ b/test-files/indentation-reference-document.ts
@@ -299,6 +299,24 @@ function bif2(a: number,
     return 1;
 }
 
+// Array shorthand.
+function bif3(a: number,
+              b: number): number[] {
+    return [1];
+}
+
+// Array shorthand in union.
+function bif4(a: number,
+              b: number): number[] | number {
+    return [1];
+}
+
+// Array shorthand in union, with spaces.
+function bif5(a: number,
+              b: number): number[   ] | number {
+    return [1];
+}
+
 // Comment where the return type would appear.
 function gogo(a: number,
               b: number) /* foo */ {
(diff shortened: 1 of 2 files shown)


josteink:
First commit looks clean and nice. Second... Seems a bit more thorough.

Giving a glance though I see the following:

* `(typescript--backward-syntactic-ws)` seems to be called in all cases anyway, so lets just call it upfront.
* `Check whether we are at "):"` moved ahead of `Looking at the arrow of a function definition`.
* The rest of the `cond` expression is mostly new, and the long `(unless (or (eq` mess is replaced by the much more compact `(memq (char-before) '(?\) ?} ?\" ?\]))`.
* `(cl-return-from search-loop nil)` being made part of the `cond`-expression since everything else was soaked up there too.

From what I can tell, there's now less code, less nesting, fewer branches/codepaths and less intermediate state. Basically an overall reduction in complexity.

As long as the tests still run, I'm happy with these changes.

I'm afraid though that I can't help and offer any improvements on this code. That would require me taking a deeper dive into the code, and unfortunately diving into code of this complexity is something I don't have much time for these days :(

lddubeau:
Thanks. I just wanted to give an opportunity to take a look, if desired, before I commit. I'm not expecting anyone to sit down and do a line by line walkthrough. Though if someone had volunteered for that, that would be fine too. :smile:

I'm going to go ahead and merge.

PR #61, opened 2017-12-12
Make the comment-start and comment-end variables local.

lddubeau:
Leaving them global as we did was a bug, which did cause me problems. I'd try to use `comment-region` in a TypeScript buffer and get my lines prepended with `# ` instead of `// ` (probably because something else was resetting the global `comment-start` and `comment-end` at some point). That was very annoying. Conversely, I ran into the reverse situation a couple times: commenting in a buffer where `# ` should have been used, resulted in `// ` being used instead. Most of the time I did not want to stop my work to track down the problem --- I just restarted the mode instead --- but I finally took the time and here's the fix.

I've also replaced the verbose code that sets local variables with `setq-local`.

This is not something I'd test for in the test suite but maybe there's an argument to be made for it??

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2453,27 +2453,22 @@ Key bindings:
   :group 'typescript
   :syntax-table typescript-mode-syntax-table
 
-  (set (make-local-variable 'indent-line-function) 'typescript-indent-line)
-  (set (make-local-variable 'beginning-of-defun-function)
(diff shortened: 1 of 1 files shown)


josteink:
Yeah. That indeed looks like a bug. Good catch. Nice cleaning up the syntax all over the place while you're at it.

LGTM. Merge ahead.

PR #62, opened 2017-12-12
Fix regexp fontification.

### LINKED ISSUE #44: font highlighter ends regexp prematurely

lddubeau:
Might as well copy here the 2nd commit message...

Use syntax-propertize-function to fontify regexps.  The complexity of regular expressions taxes the basic facilities for funtification. We reuse the latest js.el code that handles fontification to fix lingering issues with how typescript-mode has handled regexp fontification so far.

We do not simply set `syntax-propertize-function` to `js-syntax-propertize` because the latest version of js.el shipped with Emacs has a buggy implementation. Our choices are: copy from the latest version of js.el, come up with some other implementation, or accept to live with buggy fontification.

Closes #44.

I copied the functions from `js.el` wholesale and renamed them. I'm not a fan of copy-rename, but I see this as the least objectionable option from the three I mentioned above. The current bugs with regexp fontification are extremely annoying, and I don't have time to come up with my own implementation. (I'd probably end up with something similar to what `js.el` does, anyway.)

This is a significant change. So there should be discussion before it is merged. I'm not looking for a line-by-line review but there should at least be some sort of consensus.

### DIFF
--- a/typescript-mode-tests.el
+++ b/typescript-mode-tests.el
@@ -197,6 +197,9 @@ LOCATION can be either a single location, or list of locations,
 that are all expected to have the same face."
   (test-with-temp-buffer
    contents
+   ;; Make sure our propertize function has been applied to the whole
+   ;; buffer.
+   (syntax-propertize (point-max))
    (dolist (spec expected)
      (if (listp (car spec))
          (dolist (source (car spec))
@@ -289,6 +292,18 @@ declare function declareFunctionDefn(x3: xty3, y3: yty3): ret3;"
   (font-lock-test "var a = /flip\\/flop/;"
                   '(("=" . nil)
                     (("/flip" "\\\\" "/" "flop/") . font-lock-string-face)
+                    (";" . nil)))
+  ;; Make sure a forward slash in a character class is handled fine.
+  ;; It must not terminate the regular expression.
+  (font-lock-test "var a = /[/]/;"
+                  '(("=" . nil)
+                    (("/" "\\[/" "\\]/") . font-lock-string-face)
+                    (";" . nil)))
+  ;; Make sure an open bracket in a character class does not
+  ;; throw off fontification.
+  (font-lock-test "var a = /[[]/;"
+                  '(("=" . nil)
+                    (("/" "\\[\\[\\]" "/") . font-lock-string-face)
                     (";" . nil)))
   ;; A sequence of two forward slashes is never a regex, so there is
   ;; no such thing as an \"empty regex\" when we use the forward slash
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -1880,21 +1880,73 @@ This performs fontification according to `typescript--class-styles'."
                                    typescript--font-lock-keywords-3)
   "Font lock keywords for `typescript-mode'.  See `font-lock-keywords'.")
 
-;; XXX: typescript can continue a regexp literal across lines so long
-;; as the newline is escaped with \. Account for that in the regexp
-;; below.
-(defconst typescript--regexp-literal
-  "[=(,:]\\(?:\\s-\\|\n\\)*\\(/\\)\\(?:\\\\.\\|[^/]\\)+\\(/\\)"
-  "Regexp matching a typescript regular expression literal.
-Match groups 1 and 2 are the characters forming the beginning and
-end of the literal.")
+;;; Propertize
 
-;; we want to match regular expressions only at the beginning of
-;; expressions
-(defconst typescript-font-lock-syntactic-keywords
-  `((,typescript--regexp-literal (1 "|") (2 "|")))
-  "Syntactic font lock keywords matching regexps in typescript.
-See `font-lock-keywords'.")
+;;
+;; The propertize code was adapted from:
+;; https://github.com/emacs-mirror/emacs/blob/489d6466372f488adc53897435fff290394b62f7/lisp/progmodes/js.el
+;;
+
+(defconst typescript--syntax-propertize-regexp-regexp
+  (rx
+   ;; Start of regexp.
+   "/"
+   (0+ (or
+        ;; Match characters outside of a character class.
+        (not (any ?\[ ?/ ?\\))
+        ;; Match backslash quoted characters.
+        (and "\\" not-newline)
+        ;; Match character class.
+        (and
+         "["
+         (0+ (or
+              (not (any ?\] ?\\))
+              (and "\\" not-newline)))
+         "]")))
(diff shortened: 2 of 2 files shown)


josteink:
Getting this stuff fixed is nice. Getting this stuff fixed without ending up with even more complex regexp'es is even nicer.

That said... If I'm going to be devils advocate here, it will have to be that we introduce a `syntax-propertize` function.

While I have mostly positive experiences using this hook in `csharp-mode`, I've had [discussions](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=23901) with the Emacs and `cc-mode` maintainers about it in the past.

Key quote from this discussion:

> The backslash is not getting marked with a syntax-table text property.  The reason for this is, I think, that `csharp-mode` is trying to use `syntax-propertize-function` to do this.  The problem is that nothing is
triggering a call to this function.  This is one of the reasons that the CC Mode core doesn't use the `syntax-propertize-function` mechanism - it is fairly random when and whether it gets run.  Officially it describes itself as a "lazy" function, as if that were a positive aspect.

I don't remember how, but in the end we worked it out and I can't recall us having issues with `syntax-propertize-function` since then.

So the Emacs maintainers are weary about using `syntax-propertize-function` for populating syntax-properties... I guess this is fine for now, but if we expand this usage and experience syntax-table related issues in the future, we may have ourselves to blame :)

lddubeau:
Discussing issues like this one is exactly what I was hoping for.

I'm sure there *are* issues with relying on the results of `syntax-propertize-function`. Working on indentation code, I've often thought it would be nice to reuse what the fontification code does. After all, it is identifying function names, parameters, etc. I could just rely on what it has found to determine indentation. Then I realized that I could not do that. The fonctification information necessary may either be absent or stale when indentation is being performed, which would result in incorrect indentation.

However, I don't think that relying on `syntax-propertize-function` is any worse than relying on a regexp passed to the font locking mechanism. Looking at documentation, it seems to me that both are applied by the font locking machinery. Oh, and the current implementation uses `font-lock-syntactic-keywords`. The 5th element in the value assigned to `font-lock-defaults`:

```
  (setq-local font-lock-defaults
       (list typescript--font-lock-keywords
	     nil nil nil nil
	     '(font-lock-syntactic-keywords
               . typescript-font-lock-syntactic-keywords)))
```

It ends up assigned to `font-lock-syntactic-keywords`. I went back to look at its documentation, and rediscovered something I saw a while back but forgot:

> This variable is obsolete since 24.1;
>  use ‘syntax-propertize-function’ instead.
> This variable may be risky if used as a file-local variable.

`typescript-mode` is currently using the ancestor of `syntax-propertize-function`, and based on what I'm seeing, whatever pitfalls we may fall into by relying on the results of `syntax-propertize-function`, these pitfalls are *already* present due to the use of `font-lock-syntactic-keywords`. In either case, the syntactic properties of a JavaScript regexp literal are not set until the font locking mechanism has looked at the literal.

----

I was also going to say that I don't think we can write a regexp that will properly handle all valid JavaScript regexp literals... but I'm not sure whether this is true. I thought I had an example of an insurmountable hurdle, but I managed to prove that it is not unsurmountable after all. But this moot in light of the above.

josteink:
> typescript-mode is currently using the ancestor of syntax-propertize-function, and based on what I'm seeing, whatever pitfalls we may fall into by relying on the results of syntax-propertize-function, these pitfalls are already present due to the use of font-lock-syntactic-keywords

So it's confirmed that whatever change we make here, shouldn't impact fontification and syntax-properties for the worse. That's good.

Then (not even as devil's advocate) I can't see any issues with this PR.  

> I was also going to say that I don't think we can write a regexp that will properly handle all valid JavaScript regexp literals... 

A regexp to handle all regexpes.... I'm sure it can be done, but it will just as surely be evil. Very evil. Sauron evil.

If we're not worse of uses syntax-propertizing functions, I see no reason to walk further into mordor :)

As for your original PR comment:

> I copied the functions from `js.el` wholesale and renamed them. I'm not a fan of copy-rename, but I see this as the least objectionable option from the three I mentioned above.

I agree. It may not be elegant, but it's from what I can se, it's absolutely the most practical and pragmatic way to solve this.

No objections on this approach on my part.

lddubeau:
There's a bug. :frowning_face:  If I have something like this:

```
/*
 Super long line that will need to be broken into multiple pieces when the paragraph is filled.
*/
```

And I perform `fill-paragraph`, then the lines that are modified are not fontified properly. The are refontified correctly if I further edit them (e.g. space-backspace). I've performed a test with `js.el` and the problem exists there too. I've tried a bunch of things to see if I could fix it, but I've not been successful. 

I'm running out of ideas and time to spend on this. 

There's a possibility that the bug is not in the lisp code of the mode but in Emacs itself. When I have time, I may try installing newer versions than the one that comes with Debian stable and see if I can reproduce the issue there. One thing that is really strange is that if I just add a newline manually, then there is no problem. Another thing I noticed while doing this:

* When adding a newline manually, `typescript-syntax-propertize` is called *once* with a `start` and `end` that cover the whole comment.

* When filling the paragraph, `typescript-syntax-propertize` is called *twice* and the first value for `start` is after the initial `/*`.

josteink:
To verify things like this I usually run Emacs built from git master. On Debian-based distros getting this built takes surprisingly little effort and time.

lddubeau:
Progress!

I noticed that the code inside `syntax-propertize-rules` that should deal with regexps was never run. When I removed the early call to `typescript-syntax-propertize-regexp` the problem I described earlier ceased to occur. I suspect that when playing with the buffer state outside `syntax-propertize-rules`, the code needs to be more careful about what it does, whereas `syntax-propertize-rules` provides a safety harness of sorts.

I've added a commit that shows exactly what I changed.

josteink:
I'm not going to say I'm fully fluent in what's going on here, but on the surface at least, these changes look absolutely reasonable.

lddubeau:
I've used this quite a bit now and have not run into any issues.

josteink:
Great. That’s a good a test as any I can think of :)

PR #63, opened 2017-12-28
Fix the incorrect indentation of functions that have numbers in their return type annotation.

lddubeau:
This could potentially raise some organizational concerns with the new constant, so I'm submitting as a PR rather than just push to `master` to give folks a chance to comment.


josteink:
LGTM. No objections on my part.

PR #64, opened 2018-01-17
Mark typescript indent variables as safe for file local customizations

dakra:


josteink:

Looks good to me. Unless there’s any other objections, I think this should be good to merge as is.

josteink:
So no objections and build is green. Consider it merged!

Feel free to issue more patches in the future if you have anything else you feel need to be fixed :)

PR #66, opened 2018-02-16
Add option to support indented case-statements. Fix #23 

### LINKED ISSUE #23: Indentation of `case` differs between typescript and tsserver's format
typescript.el does not indent case statements in a switch, while tsserver, and by extension Tide, do indent it. And even though I prefer the not indented version, it's really awkward when the major mode and the official formatter disagree.

jack-williams:
This is a reasonably high-impact change so would appreciate any and all feedback/criticism.

Fix #23


jack-williams:
The test functions could probably be abstracted into a single function that takes test a test file and some pre and post configuration.

typescript-mode-tests.el
```diff
@@ -29,6 +29,29 @@
 
     (kill-buffer buffer)))
 
+(ert-deftest switch-case-indent-default ()
+  (let* ((buffer (find-file "test-files/switch-case-indent-default.ts")))
+    ;; double ensure mode is active
+    (typescript-mode)
+    (setq typescript-switch-indent-offset 0)
+    (let ((test-reference (typescript-test-get-doc)))
```
josteink: `typescript-mode` does not use lexcical scoping, so this `setq` could be put into the `let` block.

typescript-mode-tests.el
```diff
@@ -29,6 +29,29 @@
 
     (kill-buffer buffer)))
 
+(ert-deftest switch-case-indent-default ()
+  (let* ((buffer (find-file "test-files/switch-case-indent-default.ts")))
```
josteink: A regular `let` will be fine here. (And I see wee have one of these in our existing tests already).

typescript-mode-tests.el
```diff
@@ -29,6 +29,29 @@
...
+    (typescript-mode)
+    (setq typescript-switch-indent-offset 0)
+    (let ((test-reference (typescript-test-get-doc)))
+      (typescript-test-indent-all)
+      (should (string-equal test-reference
+                            (typescript-test-get-doc))))
+    (kill-buffer buffer)))
+
+(ert-deftest switch-case-indent-custom ()
+  (let* ((buffer (find-file "test-files/switch-case-indent.ts")))
```
josteink: A regular `let` will be fine here too.

typescript-mode.el
```diff
@@ -591,6 +591,13 @@ The value must be no less than minus `typescript-indent-level'."
   :safe 'integerp
   :group 'typescript)
 
+(defcustom typescript-switch-indent-offset 0
```
josteink: I think a good default would be matching `tsserver` instead of effectively disabling this new rule. Those who want different formatting than `tsserver` should IMO the ones who needs to go customizing.

typescript-mode.el
```diff
@@ -2214,13 +2221,22 @@ moved on success."
...
+                          (indent
+                           (cond (same-indent-p
+                                  (current-column))
+                                 (continued-expr-p
+                                  (+ (current-column) (* 2 typescript-indent-level)
+                                     typescript-expr-indent-offset))
+                                 (t
+                                  (+ (current-column) typescript-indent-level)))))
+                     (if in-switch-p
+                         (+ indent typescript-switch-indent-offset)
```
josteink: IMO this setting could be a bool rather than a number.

Example: If I dynamically customize `typescript-indent-level` (for instance through editor-config, tsconfig.json, tide-hooks, or whatever), I would expect thath setting also to apply to switch/case blocks. With the current code, it won't because this setting is a number...

If we instead just make it a boolean we can switch on or off, then we can use `typescript-indent-level` when the flag is enabled.

josteink:
Hey there and thanks for the PR!

Sorry about the slow response on my part, but now I've looked over it and I'm mostly happy.

I've left some comments on the code itself, but from a high-level point of view, I think they can be summed up as:

* use let-blocks for temporary settings in test
* make the indentation-customization variable a boolean/flag to enable/disable the behaviour, instead of a number.
* make `tsserver`-like formatting the default settings for this flag.

With those things addressed, I'll be happy to merge.

jack-williams:
Thanks for the comments. Making switch indent be a flag is a much better way of doing things.

- [x] use `let` rather than `let*` in tests.
- [x] change indent customisation to a boolean flag.
- [x] enable `tsserver` behaviour be default

josteink:
Yeah. I like that much better. I'm happy now!

Unless there's any objections from @lddubeau or @ananthakumaran within a reasonable time, I'll try to get his merged.

typescript-mode.el
```diff
@@ -591,6 +591,13 @@ The value must be no less than minus `typescript-indent-level'."
   :safe 'integerp
   :group 'typescript)
 
+(defcustom typescript-indent-switch-clauses t
+  "Enable indenting of switch case and default clauses to
+replicate tsserver behaviour. Indent level is taken to be
+`typestrict-indent-level'."
```
lddubeau: typestrict??

typescript-mode.el
```diff
@@ -2203,8 +2210,8 @@ moved on success."
           ((eq (char-after) ?#) 0)
           ((save-excursion (typescript--beginning-of-macro)) 4)
           ((nth 1 parse-status)
-           (let ((same-indent-p (looking-at
-                                 "[]})]\\|\\_<case\\_>\\|\\_<default\\_>"))
+           (let ((same-indent-p (looking-at "[]})]"))
+                 (switch-keyword-p (looking-at "default\\_>\\|case\\_>[^:]"))
```
lddubeau: Any reason we don't have `\\_<` in front of each keyword, like we do with `(looking-at "\\_<switch\\_>")` below?

lddubeau:
@jack-williams Overall it looks good. Thank you. I left a couple comments on the code. When checking the code, it looked to me like this is now handled properly:

```
{
    case: 1,
    default: 2
}
```

In previous versions both properties would be indented at the same level as the braces. It seems to me we should have a formal test for it in the indentation testing so that we don't break it again. (Highlighting is still wrong but that's a separate matter.) Because of the on/off option, I would see the test being added to both `switch-case-indent-default.ts` and `switch-case-indent-disabled.ts`

typescript-mode.el
jack-williams: I carried it over from js-mode, but it seems to work in spite of removing `\\_<`, not because of it. I'll add them back on.

jack-williams:
@josteink @lddubeau Thanks for the comments.

I'll correct the typo and add the regex symbol prefix back on.

Now that the default is to indent, should I remove `switch-case-indent-default.ts` and just uncomment the relevant region in `indentation-reference-document.ts`? We'd still want `switch-case-indent-disabled.ts`, also extended with the object literal example.

lddubeau:
@jack-williams I'm a bit concerned about the following possibility: at some point someone changes the indentation code and introduces a bug that messes up indentation that *should* not be affected by the state `typescript-indent-switch-clauses`, but the bug is revealed only when the switch is in one of its possible values. The indentation logic is hairy enough that this is more than a freak possibility.

To prevent this, code that should be indented identically no matter how `typescript-indent-switch-clauses` is set should be tested with all possible values of `typescript-indent-switch-clauses`. So maybe this:

* modify `indentation-reference-document.ts` to contain only the cases that don't depend on the value of `typescript-indent-switch-clauses`. (Add the case of an object with `case` and `default` fields here.) And modify the test suite to indent this file with both values of `typescript-indent-switch-clauses`.

* and keep `switch-case-indent-default.ts` and `switch-case-indent-disabled.ts` as they are.

jack-williams:
@lddubeau Agreed.
- Object case added to `indentation-reference-document.ts`. This should be invariant of `typescript-indent-switch-clauses`.
- Object cases added for `switch-case-indent-default.ts` and `switch-case-indent-disabled.ts`, testing objects outside a switch, in a case clause, and in a default clause.

josteink:
With the recent changes, I feel there's little left to improve.

Any objections from anyone before I merge?

lddubeau:
It looks like this proposal of mine did not make it into the PR:

> And modify the test suite to indent this file [i.e. `indentation-reference-document.ts`] with both values of `typescript-indent-switch-clauses`.

Is there an objection to this? Or did I miss something along the way?

jack-williams:
@lddubeau I misread that you wrote, sorry. I've added the test now.

lddubeau:
Merged! Thank you very much!

jack-williams:
No problem! Thanks for the comments and help @josteink @lddubeau

PR #68, opened 2018-03-14
Typo in the readme

GeertVL:


josteink:
Indeed. That's not right.

Thanks for fixing it up!

PR #69, opened 2018-03-15
Fontify yield and yield* as keyword

### LINKED ISSUE #67: font-lock yield keyword
`yield` and `yield*` are not highlighted as keywords. It's an easy workaround (below), but should probably be in this package

```elisp
(font-lock-add-keywords 'typescript-mode
                        '(("\\<\\(yield\\(\\*\\|\\>\\)\\)" . 'font-lock-keyword-face)))
```

felipeochoa:
Closes #67

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -290,7 +290,7 @@ Match group 1 is the name of the macro.")
      "private" "protected" "public" "readonly" "return" "set" "static" "string"
      "super" "switch"  "this" "throw" "true"
      "try" "type" "typeof" "var" "void"
-     "while" ))
+     "while"))                  ; yield is handled separately
   "Regexp matching any typescript keyword.")
 
 (defconst typescript--basic-type-re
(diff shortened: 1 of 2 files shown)


typescript-mode.el
```diff
@@ -318,6 +318,7 @@ Match group 1 is the name of the macro.")
                 (list "\\_<for\\_>"
                       "\\s-+\\(each\\)\\_>" nil nil
                       (list 1 'font-lock-keyword-face))
+                (cons "\\_<yield\\*?" 'font-lock-keyword-face)
                 (cons typescript--basic-type-re font-lock-type-face)
```
josteink: This lacks a closing `\\_>`, or else "yield" in "**yield**er" too will be fontified as a keyword.

josteink:

All in all it looks mostly good and I especially appreciate you adding a test-case for it too. That said, there's a small error there, which I've left a comment for.

If you could create a negative test-case for that one, and fix the regexp, I'd be more than happy to merge this!

felipeochoa:
Ok updated!

josteink:
Consider it merged! I simplified the regexp a little, but otherwise I kept things as is.

Thanks for the patch.

josteink:
It looks like my simplification broke parts of your tests without me noticing before submitting: https://github.com/ananthakumaran/typescript.el/commit/526f307ca8ca50b8f98975f2812198212f6acf2f#diff-53b8a294602753ef50f7c7d7902490d5R362

I'm not quite sure I understand *how* that test should work, as such, I'm hesitant to rush a quick fix. Care to help me out? :)

PR #70, opened 2018-03-20
Add function to convert plain strings to template strings.

lddubeau:
This is not yet ready for merging. I'm opening a PR so that the functionality can be discussed early.

This basically adds a function that allows the user to turn a non-template string into a template string. I often find myself starting to write `new Error("unexpected flag ` then while I'm writing it, I realize that I want to interpolate a variable in there so I end up with ``new Error("unexpected flag ${flag}`)`` and then I have to go back to fix the start of the string. I've been using the function I introduced here so that I can just terminate the string with the same delimiter I used to open it and then run the function to turn it into a template.

One thing I'd find useful is a key binding for this function. Not sure however what the prevailing wisdom is about assigning key bindings for this mode. (Maybe I missed the memo?)

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -631,6 +631,41 @@ seldom use, either globally or on a per-buffer basis."
   :type 'hook
   :group 'typescript)
 
+;;; Public utilities
+
+(defun typescript--move-to-end-of-plain-string ()
+  "If in a plain string, move to the end of it. A plain string is a
+string which is not a template string. Otherwise, the point does not change.
+
+Returns point if the point was changed, or nil if point was not changed.
+"
+  (let* ((syntax (syntax-ppss))
+         (str-terminator (nth 3 syntax))
+         (string-start (nth 8 syntax)))
+    (when (and string-start
+               (not (eq str-terminator ?`)))
+      ;; We just search forward and then check if the hit we get has a
+      ;; string-start equal to ours.
+      (when (re-search-forward
+             (concat "\\(?:[^\\]\\|^\\)\\(" (string str-terminator) "\\)")
+             nil t)
+        (goto-char (match-beginning 1))))))
+
+(defun typescript-convert-to-template ()
+  "Convert the string at point to a template string."
+  (interactive)
+  (save-excursion
+    (let* ((syntax (syntax-ppss))
+           (string-start (nth 8 syntax)))
+      (when (and string-start
+                 (typescript--move-to-end-of-plain-string))
+        (let ((end-start (or (nth 8 (syntax-ppss)) -1)))
+          (when (=  end-start string-start)
+            (replace-match "`" t t nil 1)))
+        (goto-char string-start)
+        (delete-char 1)
+        (insert "`")))))
+
 ;;; KeyMap
 
 (defvar typescript-mode-map


josteink:
> I often find myself starting to write new Error("unexpected flag then while I'm writing it, I realize that I want to interpolate a variable in there so I end up with new Error("unexpected flag ${flag}`) 

That's a fairly simple use-case. I have no problems supporting that.

> I'm opening a PR so that the functionality can be discussed early.

Looking at the implementation itself, it seems like you rely almost entirely on Emacs syntax tables and syntax-parsing. I'll have to say that is a good thing.

Seeing as this is also a interactive function for *code you're currently editing* I don't think there's any risk of the syntax-properties not yet being properly populated and initialized.

I'm sure I can nitpick a little here and there, but the technique used looks fine.

> One thing I'd find useful is a key binding for this function. 

I typically assign global key-bindings using a `C-x` and all mode-specific keyb-indings to a `C-c` prefix. This applies equally for custom-bindings in my  `.emacs`-file and default bindings in modes I author.

As such, I think having a `C-c` prefix is OK for such a command, but we must make sure it doesn't overlap with anything we define in `tide`, because we can safely assume most users are going to use both modes.

lddubeau:
I'll try to come up with a sensible `C-c` prefix keybinding that does not conflict with `tide`.

I was also thinking of adding an auto-check on hitting `"` and `'`. Similar to how we already have `typescript-insert-and-indent` that inserts the key and indents the line. The new command would insert the key  and check whether they character just inserted is terminating a string, and if yes, scan the string for `${...}`, and if present convert the string to a template automatically. If there's ever a false positive, a simple undo could undo the modification.

josteink:
> I'll try to come up with a sensible C-c prefix keybinding that does not conflict with tide.

Good good.

> I was also thinking of adding an auto-check on hitting " and '. Similar to how we already have typescript-insert-and-indent that inserts the key and indents the line. 

That sounds neat, but based on my experience with "smart" editors intervening when I'm writing code, I think it would be good to allow people to enable or disable this via a `defcustom`.

Sometimes these things get in your way, and you just want to be able to focus on the things you are doing, not what the editor does underneath your keypresses :)

lddubeau:
Yes, having a customization for turning on the automatic conversion to template was part of the plan. I've also made sure that the automatic conversion can be reverted with one undo step.

I've implemented the functionalities and added tests for them. It is ready for review.

josteink:
OK. I've been extremely busy the past few days and I'm likely to be even more busy the upcoming weeks (expecting our second child any day soon now)...

If I can find time tomorrow, I'll try to do a miniature review, just so that you can get *some* feedback. But if the days pass and you don't hear anything from me... Yeah. Sorry about that :)

Worst case scenario: Commit away. You're a good guy. We know that. We trust you.

lddubeau:
:tada: Congratulations re your second child. I hope the birth goes well (or went well, if it has already happened)!

And regarding committing, not committing, in general, if a change is substantial I like to at least provide the *opportunity* for discussion before committing to `master`. In the case at hand here, the one thing that I think requires discussion is my choice of `C-c '` as the keybinding for manually converting a plain string to template. It may be clashing with something I'm not aware of.

(If I feel a change of mine is so trivial that there's not much benefit for discussion, I just commit to `master`, like I did with the latest indentation fixes I committed.)

But yeah, if keeping this PR open becomes problematic, I'll just commit to `master`.

josteink:
> I hope the birth goes well (or went well, if it has already happened)!

Thanks! Still busy. Still waiting :)

> In the case at hand here, the one thing that I think requires discussion is my choice of C-c ' as the keybinding for manually converting a plain string to template. It may be clashing with something I'm not aware of.

I know I have `C-c '` (and `C-c C-'`) toggled to the `org-mode`  (or `org-babel`) command `org-edit-special`. I think that's partially a default/common setting.

Take the following org-mode buffer:

````org
* Some heading

Some text, with code-example below:

#+BEGIN_SRC typescript
  const foo = "bar";
#+END_SRC
````

Basically `org-edit-special` lets you create a dedicated buffer for the `#+BEGIN_SRC mode` section... and edit it in its native mode. You then typically re-apply this key-command inside this new buffer to exit back to the original `org-mode` document and update it with the new contents.

So you're kind having a collision with usage of typescript inside `org-mode` and `org-babel`. I'm not sure if that's a big enough deal to warrant changing the key-combo. and I must admit I'm not sure what that different key-combo should be.

josteink:
It’s not standard, no, but I’m trying to adhere to the same conventions as established in the book mastering Emacs. 

Basically it’s recommends rebinding lots of common commands to additional keys where you keep the control key consistently pressed all the time, to make them more accessible. 

Overriding this custom binding is honestly not deal-breaker, but I thought I’d mention it as it was the one collision I could think of.

PR #72, opened 2018-05-09
Remove logic handling preprocessor macros

Wilfred:
TypeScript does not have preprocessor macros. It was proposed in https://github.com/Microsoft/TypeScript/issues/4691 but never added to the language.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -38,7 +38,7 @@
 ;; The main features of this typescript mode are syntactic
 ;; highlighting (enabled with `font-lock-mode' or
 ;; `global-font-lock-mode'), automatic indentation and filling of
-;; comments and C preprocessor fontification
+;; comments.
 ;;
 ;;
 ;; General Remarks:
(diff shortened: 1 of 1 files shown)


josteink:
Thanks for another PR @Wilfred!

I'm all for removing pointless legacy code to keep the rest of the code easier to maintain.

Glancing over this, I cannot see it covered by any tests, and all the code in question seems to stem from the initial commit (meaning none of the current maintainers feel any ownership to it).

As long as everything still works, I'm supportive of these changes. Anyone else have any opinions?

josteink:
LGTM. I've just ran and tested this on some of my projects, and I can't see it introducing any regressions.

Consider it merged!

quirinpa:
Oh my, what destructive joy. I joke. You know what this would be useful for?
Conditional static imports. Oh well. Good thing we can still run gcc before tsc, or.. Use gcc for dependency resolution, even..

Turtle master, into the sky, on a cloud.

quirinpa:
Oh oh. I forgot to say: Conditional static imports. Just joking.

Remember what you have destroyed.

I'll just stop now.

josteink:
Maybe this is a joke, and I'm just slow (it's monday morning after all)...

But would you care to elaborate on this with an example? 😄

quirinpa:
You're not slow, I'm just confusing, and confused. All the time.

Here it goes;

In a config.h file (or though gcc's -D flag):
```c
#define DEVELOPMENT 1
```
And then...
```js+c
#if DEVELOPMENT
import { a } from 'whatever-dev';
#else
import { a } from 'whatever-prod';
#endif
```
It could be used for many other things as well, of course.
Kawabunga.

quirinpa:
I'm not sure if that should be typescript's responsibility...

But it would be a shame to have to let go of [Tsuquyomi](https://github.com/Quramy/tsuquyomi) in order to use the gcc preprocessor.

I wish JS build tools integrated well with oldschool tools... I assume it is hard to do single-file compilation efficiently in JS...

Dead turtle.

Wilfred:
This PR did not actually change typescript, it only changed syntax highlighting in Emacs.

PR #76, opened 2018-06-05
Drop typescript-parent-mode.

### LINKED ISSUE #75: Mode is not considered to derive from prog-mode
I was trying to use `helm-imenu-in-all-buffers` which is a helm command which combines the imenu output of all buffers with the same mode, but part of the filtering process is to ensure that the buffer derives from `prog-mode`, so the resulting list is empty because `typescript-mode` is not considered to derive from `prog-mode`.

I noticed that this mode has an interesting way of defining the parent mode by aliasing either `prog-mode` or `fundamental-mode` to `typescript-parent-mode`:

https://github.com/ananthakumaran/typescript.el/blob/7a5c74d88e3c5513cc4431a837003736f905a75e/typescript-mode.el#L2588-L2593

I would have hoped that this would still mean that `(derived-mode-p 'prog-mode)` would be true, but it doesn't. Instead, `(derived-mode-p 'typescript-parent-mode)` yields true, although that's of little use.

Is there any chance we can explicitly set it to `prog-mode`, even if by a user-defined opt-in option? I think it would be desirable because for better or worse, a lot of packages and configuration seem to check that a buffer's mode is `prog-mode` derived, which can lead to functionality breaking in subtle ways.

Reference: emacs-helm/helm#2052

lddubeau:
Putting this as a PR so that the PR checks run to make sure I did not break something.

Fixes #75.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2585,11 +2585,8 @@ the broken-down class name of the item to insert."
 
 ;;; Main Function
 
-(defalias 'typescript-parent-mode
-  (if (fboundp 'prog-mode) 'prog-mode 'fundamental-mode))
-
 ;;;###autoload
-(define-derived-mode typescript-mode typescript-parent-mode "typescript"
+(define-derived-mode typescript-mode prog-mode "typescript"
   "Major mode for editing typescript.
 
 Key bindings:


josteink:
LGTM.

Issue #39, opened 2017-06-06
Use prettier for formatting?


nspaeth:
Now that prettier/prettier#13 supports typescript, would it make sense to offload formatting to that? It seems to be gaining traction as a defacto standard.

josteink:
If you want pretty formatting, the recommended approach is to use [tide](https://github.com/ananthakumaran/tide).

By default, tide uses `tsserver` to format the code on save, giving you the standard formatting which TypeScript-code should adhere to by convention.

Is there anything else prettier could give us, that this doesn't already provide?

josteink:
To add to my former comment, I've never heard of prettier before now, and I honestly would like to keep the list of dependencies for Emacs typescript-development as short as possible.

Right now a "full stack" requires the following:

- Emacs
- typescript.el
- tide (which bundles tsserver)
- nodejs (to run tsserver)

Despite this list of dependencies, and to my great surprise, this seems to work pretty seamless, consistent and well across all platforms I've tested it (even on Windows!).

I'd like to keep it like that, and adding more tools to the chain really has to do *wonderful* things to justify the risk :)

lddubeau:
I'm going to close this. As usual, if there's a substantial argument to be made in favor of the feature proposed here, we can reopen.

lddubeau:
I've tried prettier earlier today and immediately uninstalled it. 

I have a few reflections about "support for prettier" in the context of this mode.

First, right now, it looks like if anyone wants support for it, **they'll have to step up to the plate themselves** and provide a quality PR (substantially tested) for it.

I can see two ways "support for prettier" could potentially apply to this mode:

1. Support for running prettier on files automatically or manually. **In my opinion this functionality does not belong to this mode.** That's general Emacs "run-on-save" functionality, or something to trigger through linter runs. Linter runs are handled by `flycheck`, not this mode.

2. Modify the indentation code of this mode so that its indentation naturally matches the one that prettier wants. This is where I suppose an argument could be made that `typescript-mode` should have a configuration option allowing the selection of different indentation algorithms. As I said above, someone would have to step up to the plate to make this reality.

PR #78, opened 2018-07-29
Add typescript faces

### LINKED ISSUE #32: How to enable colors in typescript
Hi, in js2-mode I see defface variables
```
(defface js2-function-call
  '((t :inherit default))
  "Face used to highlight function name in calls."
  :group 'js2-mode)
``` 
and in my themes file I added
```(js2-function-call (:foreground ,green))```
to make the function call green.

Is there any similar way to add my theme specific colors for typescript. Thanks.

tam5:
closes #32

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,17 @@ To install typescript.el simply type `M-x package-install<RET>typescript-mode<RE
 
 To customize `typescript.el` just type the following: `M-x customize-group<RET>typescript<RET>`.
 
+### Faces
+
+`typescript.el` provides the following faces that inherit from `font-lock`, but can be customized to your liking.
+
+* `typescript-string-face`
+* `typescript-keyword-face`
+* `typescript-function-name-face`
+* `typescript-variable-name-face`
+* `typescript-type-face`
+* `typescript-constant-face`
+
 You can add any other customization you like to `typescript-mode-hook`
 in your `init.el` file. `typescript.el` also handles `prog-mode-hook`
 on versions of Emacs which supports it.
--- a/typescript-mode-tests.el
+++ b/typescript-mode-tests.el
@@ -322,38 +322,38 @@ when prefixed with module modifiers."
 export function exportedDefn(x1: xty1, y1: yty1): ret1 {}\n
 export default function exportedDefaultDefn(x2: xty2, y2: yty2): ret2 {}\n
 declare function declareFunctionDefn(x3: xty3, y3: yty3): ret3;"
-    '(("basicDefn" . font-lock-function-name-face)
-      ("exportedDefn" . font-lock-function-name-face)
-      ("exportedDefaultDefn" . font-lock-function-name-face)
-      ("declareFunctionDefn" . font-lock-function-name-face)
-      (("x0" "x1" "x2" "x3") . font-lock-variable-name-face)
-      (("y0" "y1" "y2" "y3") . font-lock-variable-name-face)
+    '(("basicDefn" . typescript-function-name-face)
+      ("exportedDefn" . typescript-function-name-face)
+      ("exportedDefaultDefn" . typescript-function-name-face)
+      ("declareFunctionDefn" . typescript-function-name-face)
+      (("x0" "x1" "x2" "x3") . typescript-variable-name-face)
+      (("y0" "y1" "y2" "y3") . typescript-variable-name-face)
       (("ret0" "ret1" "ret2" "ret3") . nil))))
 
 (ert-deftest font-lock/regexp ()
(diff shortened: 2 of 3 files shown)


josteink:
Hey @tam5 and thanks for the patch!

Contributions are very much welcome, so sorry about the slow response. I've been on vacation, and didn't really have much spare time until now.

From what I can tell you've updated the tests and added documentation. That's great and I really like that!

What I'm not sure about is the *implementation* of the change itself.

From what I can see, you're just rebinding all the standard Emacs-faces into buffer-local variables, and overriding their value.

While that might work, it's a kind of "hack". Wouldn't it be better to update the `typescript--font-lock-keywords` definition further up in the code[1][2] to use the new names you've defined instead?

Is there any particular reason you chose to use the approach you did? Is this conventional in other major-modes? I ask because I honestly don't know myself.

[1] https://github.com/ananthakumaran/typescript.el/blob/master/typescript-mode.el#L289-L305
[2] https://github.com/ananthakumaran/typescript.el/blob/master/typescript-mode.el#L1678-L1821

tam5:
@josteink thanks for the review!

The short answer is I chose this approach because it made the smallest impact on the existing code.

It seems like a rather minimal effort to take the other approach, so I can switch to that if you think thats better, but first I'll give the longer answer:

First of all, I am fairly new to elisp so take what I say with a grain of salt.

In #32 I did see that you offered the suggested implementation you are speaking of, and I admit that was my initial path as well. However, though the current implementation may seem a bit like a hack, I think it may actually provide a better separation of concerns.

This approach is taken by [php-mode](https://github.com/emacs-php/php-mode/blob/master/php-mode.el#L1242-L1248), and though I don't know for sure the rational, my assumption is as follows: PHP mode uses the cc-mode engine, which provides some utilities for setting the fontification, and therefore _needs_ this hack, as php-mode alone is not directly responsible for defining all of the font-lock faces in php.

From what I've seen so far, it seems fairly uncommon to add specific overides for font-lock as we are proposing to begin with, and most major modes are only adding faces for things font-lock doesn't cover. (so automatically, no, this is not conventional)

Perhaps I'm not being very clear, but what this all points to is the fact that a major mode should ideally be using font-lock when it can. The fact that we want to add a bit more control _on top_ of that seems like a separate feature, and thus can be implemented separately. Furthermore, I think this implementation allows for greater flexibilty when it comes to future changes.

josteink:
@wasamasa: You seem to know all things about Emacs and pitfalls associated with various ways of solving different kind of Emacsy problems.

Do you have any opinion on this kind of solution? Does it look OK to you?

wasamasa:
First of all, why? What is the rationale of this? All introduced faces are named the same as the original faces. The only reason I can think of is that you're dead-set on emulating a certain editor/IDE color scheme and have discovered that Emacs has a set of generic faces that are used for all languages, with no way to change the colors per language. Solving this per mode is just wrong (it doesn't scale at all, there are too many programming languages and modes for them out there, some of which are part of Emacs), better use/write a package that allows you to override faces per mode/buffer. I'm sure they do exist.

josteink:
Thanks @wasamasa! That was exactly the kind of out-of-the-box thinking I was hoping for. 

@tam5: If there's already a generalized solution to the problem, I fail to see why we should invest in creating a specialized implementation for our own major-mode.

As such, I'm now quite in favour of closing this PR unless convinced otherwise.

Besides closing an open issue (which I guess we can close anyhow as "wont fix"), are there any other good reasons you see for implementing this?

tam5:
Indeed, that makes sense.

That’s kinda what I meant by

josteink:
> The main motivation I would see in implementing this would be more for organizational purposes. If typescript-mode were to add some of its own additional faces (like #79), then you’d want to add those to your color theme anyway

At first I thought it was odd that seemingly no default `font-lock-function-call-face` existing, but checking other major-modes (javascript-mode, js2-mode, csharp-mode), it doesn't seem like a very common thing to have.

One I did find having it though was `web-mode` (which had `web-mode-function-call-face` and inherits `web-mode-function-name-face`).

To be quite honest, I'd rather be more happy to have an official `font-lock-function-call-face` to rely on, and I might just lobby `emacs-devel` for that.

As for #39, the only other thing I can see which is non-standard is the decorator thing. Technically speaking that's a function (or constructor?) though... So I'd be happy to use `font-lock-function-call-face` for that one too. Wouldn't that be OK?

> I’m ok with closing this as won’t fix.

If so... I'm inclined to do so.

Thanks for putting in an effort and trying to improve on this mode, but sometimes even the best of PRs ends up closed for a variety of reasons. Please don't be discouraged from working on #39, even though we closed this one.

Issue #77, opened 2018-07-02
Errors no longer highlighted in compilation-mode


KevinVillela:
Between TS 2.6 and 2.7 (so maybe around May 2018) they changed the format of the error messages, resulting in next-error not always working in compilation-mode.
I believe this needs to be updated: https://github.com/ananthakumaran/typescript.el/blob/a922c27e95948a13e6a72e2620ca592a744721a7/typescript-mode.el#L2529.

josteink:
Looks fairly simple.

I’ll take a stab at it once I have time the next few days, unless of course anyone beats me to it.

josteink:
Trying to reproduce (using latest typescript) I'm not able to reproduce:

![image](https://user-images.githubusercontent.com/411338/42202404-95630f6c-7e9b-11e8-86b2-18b9d203a770.png)

Could you show some example output from your compilation-mode buffer?

KevinVillela:
Sure, sorry for the delay here, I was on holiday. I think I've figured it out, looks like the error regex doesn't like double-digit columns. Here's my .ts file:
`
  const a: string = 2;
   const a: string = 2;
`
path/to/app_component.ts:1:9 - error TS2451: Cannot redeclare block-scoped variable 'a'.

1   const a: string = 2;
          ~
path/to/app_component.ts:2:10 - error TS2451: Cannot redeclare block-scoped variable 'a'.

2    const a: string = 2;

The first one is highlighted, but the second one isn't. Seems like for whatever reason, my errors separate the line and column information with a colon, while yours use a comma.

josteink:
Ok. So testing a little on my Linux machine, I see two major changes now...

1. Ansi colours in the output... Not sure how that's going to play with our regexps.
2.Which is a lot harder to notice because of the Ansi-colours... A new format for file & line!

The format of errors themselves does not seem to have changed though, and quotes or anything like that is definitely not where the regexp fails:

````bash
➜  test-tide rm -rf node_modules && npm install typescript@2.5
test-tide@1.0.0 /home/jostein/build/test-tide
└── typescript@2.5.3  extraneous

npm WARN test-tide@1.0.0 No description
npm WARN test-tide@1.0.0 No repository field.
➜  test-tide ./node_modules/.bin/tsc                          
test.ts(2,7): error TS2322: Type '2' is not assignable to type 'string'.
test.ts(2,7): error TS2451: Cannot redeclare block-scoped variable 'a'.
[...]

PR #80, opened 2018-08-02
Document error patterns

lddubeau:
@josteink Modified as discussed in #77 

The new section I've added to the README contains code that works for me.


josteink:
LGTM

PR #82, opened 2018-08-30
Distinguish type arguments vs less/greater, fixes #81

### LINKED ISSUE #81: Indentation in the absence of semicolons breaks after parameterized type definitions
```
type Whatever<T> = T

type a = number
type b = number // this is fine

type c = Whatever<number>
    type d = number // huh!
```

The last line should not be indented.

Ptival:
As explained in #81, the current indentation fails to discern type arguments like `t < a , b > ...` from comparison operators separated by commas like `t < a , b > ...`.

It is a somewhat hard problem to solve using only regular-expressions, yet, this patch solves all of the tricky examples listed in #81, many of which are, honestly, code that should not be written without parentheses!

So this patch:
- keeps all existing good behaviors
- patches most (all?) bad behaviors for reasonable code
- patches many(?) bad behaviors for unreasonable code

And I've kept it short and documented, so that it should not become a maintenance burden.

Cheers!

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2035,7 +2035,7 @@ This performs fontification according to `typescript--class-styles'."
   "These keywords cannot be variable or type names and start a new sentence.")
 
 (defconst typescript--type-vs-ternary-re
-  (concat "[?]\\|" (typescript--regexp-opt-symbol '("as" "class" "private" "public" "readonly")))
+  (concat "[?]\\|" (typescript--regexp-opt-symbol '("as" "class" "interface" "private" "public" "readonly")))
   "Keywords/Symbols that help tell apart colon for types vs ternary operators.")
 
 (defun typescript--search-backward-matching-angle-bracket-inner (depth)
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2100,30 +2100,33 @@ RE should not expect to find characters \"<\" or \">\"."
                   ;; If we made it here, we found a candidate matching opening
                   ;; angle bracket. We still need to guess whether it actually
                   ;; is one, and not a spurious less-than operator!
-                  (progn
-                    (backward-char)
-                    ;; Look backwards for the first of:
-                    ;; - one of the symbols: = < [ ]
-                    ;; - or a TypeScript keyword
-                    ;; Depending on what comes first, we can make an educated
-                    ;; guess on the nature of our ">" of interest.
-                    (typescript--re-search-backward (concat "[=<\\[]\\|" typescript--keyword-re) nil t)
-                    (or
-                     ;; If the previous keyword is "as", we were likely in a
-                     ;; type annotation.
-                     (when (looking-at (typescript--regexp-opt-symbol '("as"))) t)
-                     ;; This final check lets us distinguish between a
-                     ;; 2-argument type "t < a , b > ..." and a use of the ","
-                     ;; operator between two comparisons "t < a , b > ...".
-                     (when (looking-at "=")
-                       (backward-char)
-                       (and
-                        ;; The heuristic is: if we keep searching backwards for
-                        ;; either a "=" symbol or the "type" keyword, the first
-                        ;; one we encounter lets us conclude. Again, see:
-                        ;; https://github.com/ananthakumaran/typescript.el/issues/81
-                        (typescript--re-search-backward-ignoring-angle-brackets (concat "[=]\\|" (typescript--regexp-opt-symbol '("type"))))
-                        (when (looking-at (typescript--regexp-opt-symbol '("type"))) t)))))))))
+
+                  ;; Look backwards for the first of:
+                  ;; - one of the symbols: = :
+                  ;; - or a TypeScript keyword
+                  ;; Depending on what comes first, we can make an educated
+                  ;; guess on the nature of our ">" of interest.
+                  (typescript--re-search-backward (concat "[=:]\\|" typescript--keyword-re) nil t)
+                  (or
+                   ;; If the previous keyword is "as", definitely a type.
+                   (when (looking-at (typescript--regexp-opt-symbol '("as"))) t)
+                   ;; A colon could be either a type symbol, or a ternary
+                   ;; operator, try to guess which.
+                   (when (looking-at ":")
+                     (and
+                      (typescript--re-search-backward
+                       (concat "[?]\\|" (typescript--regexp-opt-symbol '("as" "class" "private" "public" "readonly" "type")))
+                       nil t)
+                      (not (looking-at "?"))
+                      ))
+                   ;; This final check lets us distinguish between a
+                   ;; 2-argument type "t < a , b > ..." and a use of the ","
+                   ;; operator between two comparisons "t < a , b > ...".
+                   ;; Looking back a little more lets us guess.
+                   (when (looking-at "=")
+                     (and
+                      (typescript--re-search-backward-ignoring-angle-brackets (concat "[=}]\\|" (typescript--regexp-opt-symbol '("type"))))
+                      (when (looking-at (typescript--regexp-opt-symbol '("type"))) t))))))))
          (not (and
                (looking-at "*")
                ;; Generator method (possibly using computed property).
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2124,6 +2124,8 @@ Searches specifically for any of \"=\", \"}\", and \"type\"."
                   (or
                    ;; If the previous keyword is "as", definitely a type.
                    (looking-at "\\_<as\\_>")
+                   ;; Same goes for type imports.
+                   (looking-at "\\_<import\\_>")
                    ;; A colon could be either a type symbol, or a ternary
                    ;; operator, try to guess which.
                    (and (looking-at ":")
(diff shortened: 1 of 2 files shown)


Ptival:
Oh, I did not notice that file. I will add them. Is the intent that, the file, as it is, is correctly indented, and so running the indentation script over it should be a no-op?

lddubeau:
Yes, running the indentation script over the reference file should be a no-op.

Ptival:
Ready for review!

lddubeau:
`typescript--symbols-forbidden-in-type-parameters-re` is incorrect. I quickly came up with cases where `+`, `-` and `=` appear in type parameters. The following compiles without error with the latest `tsc`. The indentation is thrown off by the appearance of `+`, `-` and `=` in type parameters (see the lines with `const moo` and `const moo2`).

I've not found other cases (with `/`, `*`, `%`, etc.) but that does not mean they don't exist.

```
// Good case

class Q<X> {
    q: string = "a"
}

type a<X> =
    Q<X>
const blah = 1;

// Problem cases

interface Something {
    a: string;
    
    b: string;

    c: -5;
}

class Fluff<X extends Something> {
}

// Example of = and - in a type parameter.
type c<X extends Something = { a: string; b: string; c: -5; more: string }>
    = Fluff<X>
    
    const moo = 1

// Example of + in a type parameter.
type d<X extends Something = { +readonly [P in keyof Something]: Something[P] }>
    = Fluff<X>
    
    const moo2 = 1
```

Ptival:
My bad indeed! I have removed this incorrect "optimization", and now the code passes all my tests, and I believe, all the ones you added (I also added them to the test file).

Ptival:
Unfortunately, my previous incorrect optimization made it so that, when we are looking for a matching angle bracket, but there's none to be found, we end up going way deep into recursion while searching backwards.

I'd like other people's input on this, but I believe that if we see one of those keywords, we are definitely crossing by a sentence starter, thus, we should stop looking for the angle bracket:
```
const
export
function
import
let
type
var
```

I have tried to convince myself of this, indeed those words cannot, at the moment, be used as variable names. But, it could be the case that, through some syntax I am not aware of, or not thinking about, those could appear in a type annotation. Let me know if you think of a counter-example.

If you believe this is correct, my last commit adds this as a way of bailing out of the search for the opening matching bracket, whenever our search encounters one of those keywords first.

All the examples in the test-file for indentation still work, and it also solved my problem in a very large file where the recursion depth would get too deep.

Ptival:
Hold off this PR a little, I have just encountered a case I had not thought of, will try and fix it shortly.

lddubeau:
@Ptival I saw you made an additional commit. Does this mean that you've taken care of the problem you saw, and it is ready for review?

Ptival:
Yes, that last commit should have fixed that. I also removed some `(backward-char)` calls that were erroneous, now that I understand better how the point is moving via the different actions. Ready for review!

typescript-mode.el
```diff
@@ -2056,6 +2089,45 @@ This performs fontification according to `typescript--class-styles'."
...
+
+                  ;; Look backwards for the first of:
+                  ;; - one of the symbols: = :
+                  ;; - or a TypeScript keyword
+                  ;; Depending on what comes first, we can make an educated
+                  ;; guess on the nature of our ">" of interest.
+                  (typescript--re-search-backward (concat "[=:]\\|" typescript--keyword-re) nil t)
+                  (or
+                   ;; If the previous keyword is "as", definitely a type.
+                   (when (looking-at (typescript--regexp-opt-symbol '("as"))) t)
```
lddubeau: Prior to your commit, all calls to `typescript--regexp-opt-symbol` are made in the initialization of global variables. And elsewhere the regexps are just constructed manually. I'd rather keep it that way and avoid the call in code that performs performance sensitive tasks. From time to time I have to use modes that are slow at indentation. I definitely notice when Emacs has to pause to indent my line.

Here you can easily replace `(typescript--regexp-opt-symbol '("as"))` with `\\_<as\\_>`

typescript-mode.el
```diff
@@ -2056,6 +2089,45 @@ This performs fontification according to `typescript--class-styles'."
...
+                   ;; A colon could be either a type symbol, or a ternary
+                   ;; operator, try to guess which.
+                   (when (looking-at ":")
+                     (and
+                      (typescript--re-search-backward
+                       (concat "[?]\\|" (typescript--regexp-opt-symbol '("as" "class" "private" "public" "readonly" "type")))
```
lddubeau: The entire `concat` can be replaced with a precomputed string. 

Also, `"type"` in this list might face the same problem as the issue I raised on `typescript--reserved-start-keywords-re`. If it was included here for the same reason it was included there, then it should be removed here too.

typescript-mode.el
```diff
@@ -2056,6 +2089,45 @@ This performs fontification according to `typescript--class-styles'."
...
+                      (not (looking-at "?"))
+                      ))
+                   ;; This final check lets us distinguish between a
+                   ;; 2-argument type "t < a , b > ..." and a use of the ","
+                   ;; operator between two comparisons "t < a , b > ...".
+                   ;; Looking back a little more lets us guess.
+                   (when (looking-at "=")
+                     (and
+                      (typescript--re-search-backward-ignoring-angle-brackets (concat "[=}]\\|" (typescript--regexp-opt-symbol '("type"))))
+                      (when (looking-at (typescript--regexp-opt-symbol '("type"))) t))))))))
```
lddubeau: The two previous lines contain trivial calls to `typescript--regexp-opt-symbol` which ought to be replace with precomputed strings.

Also, this is another case of `(when test t)` which can be reduced to `test`.

typescript-mode.el
```diff
@@ -2031,6 +2031,39 @@ This performs fontification according to `typescript--class-styles'."
   "\\(?:NaN\\|-?\\(?:0[Bb][01]+\\|0[Oo][0-7]+\\|0[Xx][0-9a-fA-F]+\\|Infinity\\|\\(?:[[:digit:]]*\\.[[:digit:]]+\\|[[:digit:]]+\\)\\(?:[Ee][+-]?[[:digit:]]+\\)?\\)\\)"
   "Regexp that matches number literals.")
 
+(defconst typescript--reserved-start-keywords-re
+    (typescript--regexp-opt-symbol '("const" "export" "function" "import" "let" "type" "var"))
+    "These keywords cannot be variable or type names and start a new sentence.")
```
lddubeau: `type` can definitely be a variable or a type name. I can compile this without any issue. Here's an [example](https://www.typescriptlang.org/play/#src=type%20type%20%3D%20number%3B%0D%0A%0D%0Aconst%20type%3A%20type%20%3D%201%3B%0D%0A) on the playground. It does not raise any compilation errors.

I tried just removing `type` and your code still worked. I suppose we could go ahead without `type` in this list. As I understand it, these are just meant to allow stopping searches early.

lddubeau:
@Ptival I see a commit. Thanks for the update. However, I see that there are 3 changes requested in my previous review that are listed as "outdated" rather than "resolved". Did you set them aside and plan to deal with them later?

Ptival:
@lddubeau Ha! No, I missed them, they were folded and I unintentionally skipped over them. Will try to address those soon!

typescript-mode.el
```diff
@@ -2031,6 +2031,39 @@ This performs fontification according to `typescript--class-styles'."
...
+
+(defun typescript--search-backward-matching-angle-bracket (depth)
```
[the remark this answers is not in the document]
Ptival: I definitely agree with this.

Does it make sense to make the inner function be locally-scoped within the exposed one? Can you achieve this in elisp by just doing something like:
```
(defun outer
  (defun inner (depth) ...)
  (inner 1))
```
?

Or is there a preferred way or doing this?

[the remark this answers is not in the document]
Ptival: Oh, would this be a bad idea in elisp, because each call would re-define the function? I'll just follow the examples from the rest of the file and define the `inner` separately.

Ptival:
Ok, i committed changes for those 3 remaining change requests!

Ptival:
I have added two edge cases that I came upon when re-indenting my whole code base. They don't add much complexity (to either the code or the execution), but have a look!

josteink:
This PR just never ends, does it? 😄 

Good job though @Ptival, helping out making typescript-mode a better place for everyone 👍 

There's a slight conflict against the test-document in master (after your each patch). Could you resolve that as well, so that the PR is mergable?

typescript-mode.el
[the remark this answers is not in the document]
lddubeau: You ended up going the way I think is best but here are a couple of observations on this question.

I would not do `defun` inside `defun` for this. `defun` is really meant for defining "top-level" functions rather than defining local functions.  You'd use something like `cl-flet` to create a local binding. (There's also `flet` but `flet` is obsolete.)

And yes each call would create the function anew, which may have performance implications.

Given that Emacs operates largely through a "private by convention" model rather than enforce privacy trough language features, I'd do it the way you settled on: with `...-inner`.

typescript-mode.el
```diff
@@ -2031,6 +2031,53 @@ This performs fontification according to `typescript--class-styles'."
...
+  ;; looking for the matching one.
+  (or (<= depth 0)
+      (and
+       ;; If we cross over a reserved start keyword, we abandon hope of finding
+       ;; a matching angle bracket.  This prevents extreme recursion depths.
+       (typescript--re-search-backward (concat "[<>]\\|" typescript--reserved-start-keywords-re) nil t)
+       (case (char-after)
+         (?< (typescript--search-backward-matching-angle-bracket-inner (- depth 1)))
+         (?> (typescript--search-backward-matching-angle-bracket-inner (+ depth 1)))
+         (otherwise nil)))))
```
lddubeau: I'm pretty sure the explicit `(otherwise nil)` is redundant. If a `case` does not match any of its branches, then `nil` is returned by default.

Ptival:
Done! :-)

lddubeau:
Thanks. It is now merged!

PR #84, opened 2018-09-06
Fix unwanted indentation after non-keyword "each"

### LINKED ISSUE #83: Indentation without semicolons breaks after method calls named "each"
```
// Good
foo.noteach()
bar()

// Bad
foo.each()
    bar()
```

I might look into it and submit a PR if I feel motivated! :-)

Ptival:
Fixes #83 

AFAICT, this simple fix is enough. I'm not sure why "each" was in this list honestly, maybe it is a leftover from an old version of TypeScript where it was a keyword, but it is definitely not a keyword as of today:

https://github.com/Microsoft/TypeScript/blob/8f654f0f1ea42258275c67a7315c198f56d161f2/src/compiler/types.ts#L121-L197

### DIFF
--- a/test-files/indentation-reference-document.ts
+++ b/test-files/indentation-reference-document.ts
@@ -495,3 +495,6 @@ function blipblop(): void {
         const q = 1;
     }
 }
+
+container.each(x => x)
+something() // No reason for this to be indented! (cf. issue #83)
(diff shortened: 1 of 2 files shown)


lddubeau:
Thanks. Historically, `typescript-mode` was forked from `js-mode`. That's where "each" came from. Doing a `git blame` on the line you changed, I see that line remained unmodified from the initial commit so that's from `js-mode`. Probably `js-mode` had it because of this:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for_each...in

PR #87, opened 2018-09-14
Add more jsdoc tags

### LINKED ISSUE #86: Support face for missing JSDoc tags
## Description
Currently, in https://github.com/ananthakumaran/typescript.el/blob/c2d72b7d816a0ede209c507b1800f98b9d53559b/typescript-mode.el#L417-L531
there are some missing tags, e.g., `@abstract`, `@async`, `@external`, and `@todo`, ...
You can see those tags in http://usejsdoc.org/.
Could you complete tags in http://usejsdoc.org/? If it's hard, could you change those lists into custom variable so that user can change them easily?

Ailrun:
Current jsdoc tags in typescript.el do not include some tags in http://usejsdoc.org/.
This PR does
1. Update jsdoc tags to match with current js2-mode
2. Add more tags to include (almost) all tags in http://usejsdoc.org/

However, there are still some missing tags, like `@see`. `{@link}`, ... which could not be simply added by just adding their name into variables.

Resolves #86.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -436,6 +436,7 @@ Match group 1 is MUMBLE.")
   "Matches jsdoc tags with optional type and optional param name.")
 
 ;; This was taken from js2-mode.
+;; and extended with tags in http://usejsdoc.org/
 (defconst typescript-jsdoc-typed-tag-regexp
   (concat typescript-jsdoc-before-tag-regexp
           "\\(@\\(?:"
@@ -459,11 +460,13 @@ Match group 1 is MUMBLE.")
   "Matches jsdoc tags with optional type.")
 
 ;; This was taken from js2-mode.
+;; and extended with tags in http://usejsdoc.org/
 (defconst typescript-jsdoc-arg-tag-regexp
   (concat typescript-jsdoc-before-tag-regexp
           "\\(@\\(?:"
           (regexp-opt
-           '("alias"
+           '("access"
+             "alias"
              "augments"
              "base"
              "borrows"
@@ -474,36 +477,51 @@ Match group 1 is MUMBLE.")
              "define"
              "emits"
              "exception"
+             "extends"
+             "external"
              "fires"
              "func"
              "function"
+             "host"
+             "kind"
+             "listens"
              "member"
-             "memberOf"
+             "memberof"
              "method"
+             "mixes"
              "module"
              "name"
              "namespace"
+             "requires"
              "since"
              "suppress"
              "this"
              "throws"
+             "var"
+             "variation"
              "version"))
           "\\)\\)\\s-+\\([^ \t]+\\)")
   "Matches jsdoc tags with a single argument.")
 
-;; This was taken from js2-mode.
+;; This was taken from js2-mode
+;; and extended with tags in http://usejsdoc.org/
 (defconst typescript-jsdoc-empty-tag-regexp
   (concat typescript-jsdoc-before-tag-regexp
           "\\(@\\(?:"
           (regexp-opt
-           '("addon"
+           '("abstract"
+             "addon"
+             "async"
              "author"
              "class"
+             "classdesc"
              "const"
              "constant"
              "constructor"
              "constructs"
              "copyright"
+             "default"
+             "defaultvalue"
              "deprecated"
              "desc"
              "description"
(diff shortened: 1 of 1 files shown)


typescript-mode.el
```diff
@@ -422,14 +422,21 @@ Match group 1 is MUMBLE.")
...
           "\\s-*\\({[^}]+}\\)?"         ; optional type
           "\\s-*\\[?\\([[:alnum:]_$\.]+\\)?\\]?"  ; name
           "\\_>")
   "Matches jsdoc tags with optional type and optional param name.")
 
 ;; This was taken from js2-mode.
+;; and extended with tags in http://usejsdoc.org/
 (defconst typescript-jsdoc-typed-tag-regexp
   (concat typescript-jsdoc-before-tag-regexp
           "\\(@\\(?:"
```
Ailrun: Do we need these `\\(?:` part? It looks like `regexp-opt` returns string that is already wrapped in `\\(?: ... )`.

typescript-mode.el
```diff
@@ -466,67 +477,96 @@ Match group 1 is MUMBLE.")
...
              "exception"
+             "extends"
+             "external"
              "fires"
              "func"
              "function"
+             "host"
+             "kind"
+             "listens"
              "member"
```
Ailrun: `@member` takes type and name, so it should go into `typescript-jsdoc-param-tag-regexp`. Is there any problem with this modification?

typescript-mode.el
```diff
@@ -466,67 +477,96 @@ Match group 1 is MUMBLE.")
...
+             "mixes"
+             "module"
              "name"
              "namespace"
+             "requires"
              "since"
              "suppress"
              "this"
              "throws"
+             "var"
```
Ailrun: This is synonym of `"member"`, so if we move `"member"` into `typescript-jsdoc-param-tag-regexp`, this one should be moved too.

typescript-mode.el
Ailrun: Emacs wiki said it appends `\\(?:` after Emacs23. I think we can safely remove these extra groups.

josteink: If so, it might make sense to remove them, but before we do that... I’d like for us to setup some fontification tests to ensure we don’t break anything. 

Is that ok?

Ailrun: Of course.

josteink: I see we already have some tests for Jsdoc. 

As such feel free to go ahead with this modification.

typescript-mode.el
josteink: Would you be able to author a test-case for that in typescript-model-tests.el?

If so, go ahead!

typescript-mode.el
josteink: Agreed

Ailrun:
@josteink Could we merge this first? I think rearranging tags with new tests should be separated into new PR.

josteink:
>  Could we merge this first? I think rearranging tags with new tests should be separated into new PR.

Yes. That makes sense. I fully agree.

typescript-mode.el
```diff
@@ -444,18 +451,22 @@ Match group 1 is MUMBLE.")
...
              "return"
              "returns"
+             "yield"
+             "yields"
+             "type"
              "throw"
-             "throws"
-             "type"))
-          "\\)\\)\\s-*\\({[^}]+}\\)?")
+             "throws"))
```
josteink: This list used to be in alphabetical order, with "throw" before "type". With these changes it no longer is. 

Could you fix that?

typescript-mode.el
```diff
@@ -466,68 +477,97 @@ Match group 1 is MUMBLE.")
...
              "noalias"
              "noshadow"
              "notypecheck"
              "override"
+             "overview"
              "owner"
+             "package"
              "preserve"
-             "preserveTry"
+             "preservetry"
```
josteink: I see there's a few modifications of camelCased entries to be all lower-case.

Are these functional changes, or just due to source-code consistency and aesthetics?

josteink:
I have a few absolutely minor nitpicks, but apart from that this looks good.

Once those 2 things are settled, I'll be happy to merge.

typescript-mode.el
Ailrun: Oh, OK.

typescript-mode.el
Ailrun: Since [JSDoc does not care case of tags](https://github.com/jsdoc3/jsdoc/issues/1046), it should be only for consistency, since many of tags (`@defaultvalue`, `@classdesc`, ... etc.) already use lower-only case.

Ailrun:
Could you explain those minor nitpicks too? I want to know it before I send a PR for test.

josteink:
Looks good to me. Thanks for the patch!

PR #88, opened 2018-09-14
Fix indentation after keyword-lookalike methods.

### LINKED ISSUE #85: Over-indentation after call to `catch` method
```
function f1(x) {
    return (
        x.then()
    )
}

function f2(x) {
    return (
        x.catch()
            )

```

(Sorry for all the issues! :-p)

Ptival:
Proposal to fix #85 

Problem:
The original script looks for optionally-braceless constructs, and indents when it thinks it reaches them.
Unfortunately, those same keywords can also be method names.

Solution:
Distinguish a method call by looking for a period just before the keyword.

### DIFF
--- a/test-files/indentation-reference-document.ts
+++ b/test-files/indentation-reference-document.ts
@@ -627,3 +627,28 @@ function blipblop(): void {
 
 container.each(x => x)
 something() // No reason for this to be indented! (cf. issue #83)
+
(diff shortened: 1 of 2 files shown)


josteink:
Looks good to me!

As a fun note, Github also seems to highlight the keywords wrongly/inconsistently :)

PR #90, opened 2018-10-02
add `never` type

jack-williams:


josteink:
Haven’t tested yet, but looks good to me.

josteink:
Tested and works as expected. Consider it merged.

Thanks for the patch!

jack-williams:
No problem!

PR #92, opened 2018-10-16
Add autoload cookie with safe-local-var for typescript-indent-level

dakra:
If this is not autoloaded then it is only marked as safe from the
defcustom macro *after* the package is loaded.
This may be to late if someone opens a project with
`typescript-indent-level` as dir-local set but has not
loaded typescript-mode before.


josteink:
Hey there and thanks for another patch!

As for the problem itself, I'll take your word for it. It certainly doesn't seem like it can hurt, and the build is still green so I'm OK with this.

Consider it merged!

PR #95, opened 2018-11-30
Added regex constant for method calls

OlleLinderos:
It has yet to be implemented, it got reset on cache clear and I was unable to bypass it.


josteink:
I'll be happy to review a PR for this, once complete.

As of now... You just having committed a regular expression... Maybe adding some test-cases for that RX might be a good idea?

PR #96, opened 2018-12-19
Highlight multiple implements types

gexplorer:
After **implements** only the first element was highlighted. This pull requests highlights all comma separated elements.

### DIFF
--- a/typescript-mode-general-tests.el
+++ b/typescript-mode-general-tests.el
@@ -354,10 +354,12 @@ declare function declareFunctionDefn(x3: xty3, y3: yty3): ret3;"
   "Type names should be highlighted in definitions."
   ;; Typical case.
   (test-with-fontified-buffer
-      "export class Foo extends Bar implements Qux {}"
+      "export class Foo extends Bar implements Qux, Ajx {}"
     (should (eq (get-face-at "Foo") 'font-lock-type-face))
     (should (eq (get-face-at "Bar") 'font-lock-type-face))
-    (should (eq (get-face-at "Qux") 'font-lock-type-face)))
+    (should (eq (get-face-at "Qux") 'font-lock-type-face))
+    (should (eq (get-face-at "Ajx") 'font-lock-type-face))
+    )
   ;; Ensure we require symbol boundaries.
   (test-with-fontified-buffer
       "Notclass Foo"
--- a/typescript-mode-general-tests.el
+++ b/typescript-mode-general-tests.el
@@ -358,6 +358,7 @@ declare function declareFunctionDefn(x3: xty3, y3: yty3): ret3;"
     (should (eq (get-face-at "Foo") 'font-lock-type-face))
     (should (eq (get-face-at "Bar") 'font-lock-type-face))
     (should (eq (get-face-at "Qux") 'font-lock-type-face))
+    (should (eq (get-face-at ",") 'nil))
     (should (eq (get-face-at "Ajx") 'font-lock-type-face))
     )
   ;; Ensure we require symbol boundaries.
--- a/typescript-mode-general-tests.el
+++ b/typescript-mode-general-tests.el
@@ -354,13 +354,26 @@ declare function declareFunctionDefn(x3: xty3, y3: yty3): ret3;"
   "Type names should be highlighted in definitions."
   ;; Typical case.
   (test-with-fontified-buffer
+      "export class Foo extends Bar implements Qux {}"
+    (should (eq (get-face-at "Foo") 'font-lock-type-face))
+    (should (eq (get-face-at "Bar") 'font-lock-type-face))
+    (should (eq (get-face-at "Qux") 'font-lock-type-face)))
+  (test-with-fontified-buffer
       "export class Foo extends Bar implements Qux, Ajx {}"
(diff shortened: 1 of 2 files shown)


typescript-mode-general-tests.el
```diff
@@ -354,10 +354,12 @@ declare function declareFunctionDefn(x3: xty3, y3: yty3): ret3;"
...
   ;; Typical case.
   (test-with-fontified-buffer
-      "export class Foo extends Bar implements Qux {}"
+      "export class Foo extends Bar implements Qux, Ajx {}"
     (should (eq (get-face-at "Foo") 'font-lock-type-face))
     (should (eq (get-face-at "Bar") 'font-lock-type-face))
-    (should (eq (get-face-at "Qux") 'font-lock-type-face)))
+    (should (eq (get-face-at "Qux") 'font-lock-type-face))
+    (should (eq (get-face-at "Ajx") 'font-lock-type-face))
+    )
```
josteink: I appreciate the test here. That’s great!

But shouldn’t we also test the comma stuck in between as well (for not being a type-face?)

typescript-mode.el
```diff
@@ -1749,8 +1749,8 @@ and searches for the next token to be highlighted."
     (,(rx symbol-start "extends" (+ space) (group (+ (or (syntax word) (syntax symbol)))))
      (1 font-lock-type-face))
 
-    (,(rx symbol-start "implements" (+ space) (group (+ (or (syntax word) (syntax symbol)))))
-     (1 font-lock-type-face))
+    (,(rx symbol-start "implements" (+ space))
+     (,(rx symbol-start (+ (syntax word))) nil nil (0 font-lock-type-face)))
```
josteink: I’m not sure I 100% follow this code, but if the tests pass, I’m ok with it 🙂

typescript-mode-general-tests.el
lddubeau: I agree. Even if the current code works, we also want tests to protect against regressions. Regexes are easy to mess up. A future modification could fail to skip the comma.

MarcusRiemer: Just my uninvited 2 cents: Shouldn't this also test whether the basic case `Foo implements Bar` still works? Just to ensure the comma is not suddenly regarded as a required part to make the highlighting happen.

josteink: Good point. If we want to be rigorous we should probably have test-cases for 1, 2 and “many”.

typescript-mode.el
lddubeau: I went and checked the documentation of `font-lock-keywords` because that's certainly not fresh in my memory. Seems to me the test on line 1753 is a `MATCH-ANCHORED` in Emacs' parlance. Reading the doc, I see that `MATCH-ANCHORED` is scoped to a single line by default. So I thought of a probable problem with the current code. This highlights correctly (identifiers `A`, `B` and `C` are all highlighted everywhere they appear):

```
interface A {
}

interface B {
}

class C implements A, B {
}
```

However in this code, the `B` after `implements A,` is not highlighted:

```
interface A {
}

interface B {
}

class C implements A,
B {
}
```

I think that's because `MATCH-ANCHORED` operates within a line, and `B` is on another line than `implements`. The documentation for `MATCH-ANCHORED` mention that it is possible to extend the scope to subsequent lines but does not recommend it, as it may cause performance issues and other problems.

I'm not sure what the solution should be.

josteink: Thorough as always, I see. :)

I agree this represents a shortcoming, but on overall this PR still represents improvement over what we currently have. 

If a perfect solution can’t be found for now (holidays and all), I’ll be happy with this as an intermediate improvement wrt both ta-support and test-coverage.

gexplorer: > I went and checked the documentation of `font-lock-keywords` because that's certainly not fresh in my memory. Seems to me the test on line 1753 is a `MATCH-ANCHORED` in Emacs' parlance. Reading the doc, I see that `MATCH-ANCHORED` is scoped to a single line by default. So I thought of a probable problem with the current code. This highlights correctly (identifiers `A`, `B` and `C` are all highlighted everywhere they appear):
> 
> ```
> interface A {
> }
> 
> interface B {
> }
> 
> class C implements A, B {
> }
> ```
> However in this code, the `B` after `implements A,` is not highlighted:
> 
> ```
> interface A {
> }
> 
> interface B {
> }
> 
> class C implements A,
> B {
> }
> ```
> I think that's because `MATCH-ANCHORED` operates within a line, and `B` is on another line than `implements`. The documentation for `MATCH-ANCHORED` mention that it is possible to extend the scope to subsequent lines but does not recommend it, as it may cause performance issues and other problems.
> 
> I'm not sure what the solution should be.

That's what I found when trying to do something similar for **import** blocks. Even though I got it working for multiline **import** the doc says that performance could be a problem, so I'm not sure what to do.

typescript-mode-general-tests.el
gexplorer: I updated the test cases with your feedback, thanks

gexplorer:
@josteink I think that we could merge this PR and I'll submit a new one with the multi-line support.

josteink:
LGTM! Consider it merged!

Thanks for the patch. Feel free to issue more patches in the future!

PR #98, opened 2019-01-24
Provisional - Add not keyword

jack-williams:
Prepare a PR for this [feature](https://github.com/Microsoft/TypeScript/pull/29317) currently under review.


josteink:
Looks fair enough. I see no problem merging this as soon as this feature is merge to TypeScript master.

josteink:
Just letting you know, we're still checking out dangling PRs.

Unfortunately, the underlying proposal is still not settled, so I guess we'll just leave this lingering a little bit more :smile:

jack-williams:
I don’t think the proposal is moving anytime soon, so probably better to close this up and revisit if we need to!

Issue #79, opened 2018-07-29
Enable more colors in typescript


tam5:
Follow up to #32 which requests customizable faces specific to typescript mode.

This is addressed by PR #78, and allows the currently highlighted faces by typescript mode to be customized.

However, it does not address adding additional highlighting rules. In fact, the example given by @it6 in #32 is for `function-call`, which is not currently given a face at all.

I am proposing to add additional highlighting for the following cases, each with an example, and the face name it would be given:

* function calls (`typescript-function-call-face`)
```ts
    foo.bar()
        ^^^
```
* function definitions in classes (`typescript-function-name-face`)
```ts
class Foo {
    bar () {
    ^^^
      
    }
}
```
* variable names definitionitions in classes (`typescript-variable-name-face`)
```ts
class Foo {
    private foo = 'bar'
            ^^^
}
```

* decorators (`typescript-decorator-face`)
```ts
@Component({
^^^^^^^^^^
    //
})
class Foo {
    //
}
```

josteink:
All those changes sounds good to me, and if someone wants to put in the work to implement it, I will be happy to review and merge.

Right now [we have 3 font-lock levels](https://github.com/ananthakumaran/typescript.el/blob/master/typescript-mode.el#L1929-L1933). If we're adding these new ones... Maybe that should go in a fourth?

If so, I'd support making the fourth one the new default, but I know there are people who in the past have thought that to be a bit "too busy", so they should have the option to keep things as they are (that is, customize to stay on level 3 or below).

Does that sound good?

lddubeau:
Putting the additional highlighting in a 4th level seems reasonable. Though I wonder in practice whether it may make integration with other tools (minor modes, etc.) more complicated. I have no evidence one way or the other.

There's probably a segment of our user-base that would welcome the added fontification, but someone who wants to see this functionality integrated to `typescript-mode` needs to champion the change and implement it. In general, I'm in favor of adding fontification details, though I may not *personally* find them useful. But I have some reservations about some of the proposals made here.

I'm not in favor of marking `window` and `document` as builtins, because they do not always refer to the builtins. This is perfectly valid code:

```typescript
function fnord(window: number): void {
   console.log(window); // `window` is not a builtin here
}

fnord(1);
```

(Case in point, Github's fontification erroneously marks `window` as a builtin, although it is *not*!!)

[...]

PR #110, opened 2019-08-07
Enable more colors

tam5:
Addressed #79.

This pull request contains the beginnings of the work to add additional fontification to `typescript-mode`.

So far, the following changes have been implemented as a 4th level of highlighting:

- [x] function calls (`font-lock-function-name-face`)
- [x] function definitions in classes (`font-lock-function-name-face`)
- [x] decorators (`font-lock-function-name-face`)
- [x] builtins (right now just `console`, as `font-lock-type-face`)
- [x] access modifiers (`private`, `public`, etc. as `typescript-access-modifier-face` - inherits `font-lock-keyword-face`
- [x] type hints in generics (`font-lock-type-face`)
- [x] basic types (`any`, `string`, `boolean`, etc. as `typescript-primitive-face` - inherits `font-lock-keyword-face`
- [x] moved constants (false, NaN, etc) to `font-lock-variable-name-face` for more contrast (only for this 4th level)
- [x] `this` keyword (`typescript-this-face` - inherits `font-lock-keyword-face`)

Currently, the changes remain very simple and purely regexp based. This will probably have to change to properly support type hints, but I haven't yet been able to fully wrap my head around the whole `pitem` system.

Also, there are no tests are documentation or anything yet, but can add that if this gets more traction.

Remaining:
- [ ] type hints
- [ ] tests

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -300,7 +300,10 @@ Match group 1 is MUMBLE.")
 (defconst typescript--builtin-re
   (typescript--regexp-opt-symbol
    '("console"))
-  "Regular expression matching builtin stuff.")
+  "Regular expression matching builtins.")
+
+(defconst typescript--function-call-re "\\(\\w+\\)\s*("
+  "Regular expression matching function calls.")
 
 (defconst typescript--font-lock-keywords-1
   (list
@@ -619,7 +622,7 @@ Match group 1 is MUMBLE.")
 
 (defface typescript-access-modifier-face
   '((t (:inherit font-lock-keyword-face)))
-  "Face used to highlight 'this' keyword."
+  "Face used to highlight access modifiers."
   :group 'typescript)
 
 ;;; User Customization
@@ -1853,9 +1856,9 @@ and searches for the next token to be highlighted."
      (1 font-lock-type-face))
 
     ;; variable declarations
-    ;; ,(list
-    ;;   (concat "\\_<\\(const\\|var\\|let\\)\\_>\\|" typescript--basic-type-re)
-    ;;   (list #'typescript--variable-decl-matcher nil nil nil))
+    ,(list
+      (concat "\\_<\\(const\\|var\\|let\\)\\_>\\|" typescript--basic-type-re)
+      (list #'typescript--variable-decl-matcher nil nil nil))
 
     ;; class instantiation
     ,(list
@@ -1996,7 +1999,6 @@ This performs fontification according to `typescript--class-styles'."
         return t
         else do (goto-char orig-end)))
 
-(defconst typescript--function-call-re "\\(\\w+\\)\s*(")
 (defconst typescript--font-lock-keywords-4
   (append typescript--font-lock-keywords-3
     (list
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -305,7 +305,7 @@ Match group 1 is MUMBLE.")
    '("console"))
   "Regular expression matching builtins.")
 
-(defconst typescript--function-call-re "\\(\\w+\\)\s*("
+(defconst typescript--function-call-re "\\(\\w+\\)\\(<.+>\\)?\s*("
   "Regular expression matching function calls.")
 
 (defconst typescript--font-lock-keywords-1
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -1966,10 +1966,17 @@ This performs fontification according to `typescript--class-styles'."
         return t
         else do (goto-char orig-end)))
 
+(defconst typescript--function-call-re "\\(\\w+\\)\s*(")
+(defconst typescript--font-lock-keywords-4
+  (append typescript--font-lock-keywords-3
+          (list (list typescript--function-call-re 1 font-lock-function-name-face)))
+  "Level four stuff.")
+
 (defconst typescript--font-lock-keywords
-  '(typescript--font-lock-keywords-3 typescript--font-lock-keywords-1
+  '(typescript--font-lock-keywords-4 typescript--font-lock-keywords-1
                                    typescript--font-lock-keywords-2
-                                   typescript--font-lock-keywords-3)
+                                   typescript--font-lock-keywords-3
+                                   typescript--font-lock-keywords-4)
   "Font lock keywords for `typescript-mode'.  See `font-lock-keywords'.")
 
 ;;; Propertize
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -284,6 +284,10 @@ Match group 1 is MUMBLE.")
    '("any" "bool" "boolean" "bigint" "never" "number" "string" "unknown" "void"))
   "Regular expression matching any predefined type in typescript.")
 
+(defconst typescript--type-hint-re
+   (concat "\\:\\s-+" "\\(" typescript--name-re "\\)")
+   "Regular expression matching type hints that follow the `:' operator.")
+
 (defconst typescript--constant-re
   (typescript--regexp-opt-symbol '("false" "null" "undefined"
                                  "Infinity" "NaN"
@@ -306,7 +310,8 @@ Match group 1 is MUMBLE.")
                       (list 1 'font-lock-keyword-face))
                 (cons "\\_<yield\\(\\*\\|\\_>\\)" 'font-lock-keyword-face)
                 (cons typescript--basic-type-re font-lock-builtin-face)
-                (cons typescript--constant-re font-lock-type-face)))
+                (list typescript--type-hint-re 1 font-lock-type-face)
+                (cons typescript--constant-re font-lock-variable-name-face)))
   "Level two font lock keywords for `typescript-mode'.")
 
 ;; typescript--pitem is the basic building block of the lexical
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -270,19 +270,22 @@ Match group 1 is MUMBLE.")
   (typescript--regexp-opt-symbol
    '("abstract" "any" "as" "async" "await" "boolean" "bigint" "break" "case" "catch" "class" "const"
      "constructor" "continue" "declare" "default" "delete" "do" "else"
-     "enum" "export" "extends" "extern" "finally" "for"
-     "function" "from" "get" "goto" "if" "implements" "import" "in" "instanceof"
-     "interface" "keyof" "let" "module" "namespace" "new" "of"
-     "private" "protected" "public" "readonly" "return" "set" "static"
-     "super" "switch" "throw" "true"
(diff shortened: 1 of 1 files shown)


henrikrudstrom:
was just contemplating the lack of colors in typescript-mode, so you have my moral support and a bump.
also, highlighting variables inside template literals would be awesome, ie:
```
`hello ${world}`
```

josteink:
Wow. This has been lying around for 15 days untouched. That's not cool.

I'll give it a spin and see how I feel about it, and then afterwards I'll try to dig deeper into the code to see I have any issues with how things are implemented, or if I see things which could be improved.

Sorry about the delay!

josteink:
Ok so a couple of comments/impressions, just from browsing my existing code-base with and without.

* Out of the box it feels overly colourly. 
* Reverting to `master` suddenly feels very bland. Now I want colours.
* But not that much colours? Not sure.
* Some people will want it to be more bland though... This should be implemented as a [`font-lock level`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Levels-of-Font-Lock.html) people can customize, as defined in the Emacs-manual.  Oh wait, it is. Great, once again!
* Yay generics looking better! Yes!
* Does not handle multiple types, with constraints though (like `TValue` will not be highlighted in `ControlBase<TElement extends HTMLElement, TValue>`)
* I'm not sure I like having the same face for function-*definition* and function-*invocation*. And I really dislike that Emacs doesn't have any standard faces to express this semantical difference.
* Some type-keywords (like `undefined`) are suddenly highlighted with `font-lock-variable-name` instead of `font-lock-constant-face`. I'm going to consider this an error, more than an estetical disagreement.
* No tests are broken, which means we should probably improve our test-coverage.
* No additional tests are provided, which I think is a shame. We can stall writing these until we agree on the "spec" per se though.
* Still not having types highlighted feels like a big omission with all these other changes landing...  Maybe I will change my stance on these, since it's in a optional `font-lock` level?

Let me know what you think. In the mean time, I'll try to think about these changes some more, and thank you for contributing! This looks like a big patch, and we should definitely get this landed!

tam5:
Thank you @josteink for getting back to this, and thank you for taking the time to check out the code and actually run it. I very much appreciate your feedback, and I'll try and go through each point.

> - Out of the box it feels overly colourly.
> - Reverting to master suddenly feels very bland. Now I want colours.
> - But not that much colours? Not sure.
> - Some people will want it to be more bland though... This should be implemented as a font-lock level people can customize, as defined in the Emacs-manual. Oh wait, it is. Great, once again!

For the past while I've split my work machine and my home machine so one uses the new colors and the other does not. I'll agree that the initial switch in either direction is a bit jarring, but I get used to it after a couple minutes. The other thing I'll note here is the 'blandness' definitely does depend largely on the color theme being used. 

> - Yay generics looking better! Yes!
> - Does not handle multiple types, with constraints though (like TValue will not be highlighted in ControlBase<TElement extends HTMLElement, TValue>)

Good catch! I can probably add this easily.

> - I'm not sure I like having the same face for function-definition and function-invocation. And I really dislike that Emacs doesn't have any standard faces to express this semantical difference.

Firstly, I kind of agree. On the other hand, VSCode and Sublime (at least with the color themes I've checked) also use the same highlighting for these. It would be nice if Emacs had another face we could use to at least give the option of a new highlight color, but oh well for now. Lastly, the fact that they are the same actually made the implementation easier as they fall under the same regexp. To change them might be difficult and require proper contextual understanding.

> - Some type-keywords (like undefined) are suddenly highlighted with font-lock-variable-name instead of font-lock-constant-face. I'm going to consider this an error, more than an estetical disagreement.

Not sure why I did that, will revert.

> - No tests are broken, which means we should probably improve our test-coverage.

:)

> - No additional tests are provided, which I think is a shame. We can stall writing these until we agree on the "spec" per se though.

My thoughts exactly, as mentioned in the original post.

> - Still not having types highlighted feels like a big omission with all these other changes landing... Maybe I will change my stance on these, since it's in a optional font-lock level?

I totally agree that type hints are pretty crucial, just have not figured out a proper implementation yet. I'll keep thinking and maybe it will come to me eventually.

josteink:
Yeah that's a lot better. I appreciate that.

Looking more into it, I've decided that what I didn't like before was that (with my theme) these changes made my `typescript-mode` start lock almost monotone. Basically with my slightly modified `seti-theme`, `typescript-mode` became way too "blue".

But it turns out the reason for that was that not only is function-invocation and function-declaration the same font-lock face (`font-lock-function-name`), but with my theme, the same face and colour was used for `font-lock-variable-name`.

By customizing my theme to make `font-lock-variable-name` a different colour, I no longer feel like `typescript-mode` is yelling blue in my face.

Basically, at least from a visual point of view, I'm happy with these changes as they are.

That leaves us:

1. finding and fixing reported bugs (like the generics bug I mentioned in last comment)
2. code-review
3. test-cases

As you may have noticed, type-hinting is somewhat a touchy subject in `typescript-mode`, so I suggest we leave that for later, once these changes are landed.

Sounds good?

josteink:
I've been using this a little more, and I really like it. Especially with my new modified theme.

Do you plan on fixing the reported issues so far, or should I just start reviewing the code as is?

typescript-mode.el
```diff
@@ -1840,7 +1878,7 @@ and searches for the next token to be highlighted."
     ;; formal parameters
     ,(list
       (concat
-       "\\_<function\\_>\\(\\s-+" typescript--name-re "\\)?\\s-*(\\s-*"
+       "\\_<function\\_>\\|=\\(\\s-+" typescript--name-re "\\)?\\s-*(\\s-*"
```
josteink: I don't understand this change. Is this deliberate?

josteink:
I've looked over this a little more, done some more testing, and I think we're quite close to commiting this one.

That said, on top of the existing limitation (list generic arguments not being highlighted correctly, for instance `ControlBase<TElement extends HTMLElement, TValue>`), I've found a another case. Generics and methods doesn't seem to work (highlight type) when provided as argument to a generic method. Example:

````typescript
private async innerExecuteAsync<TResponse>(endpoint: string, data?: any): Promise<TResponse> {
   // do stuff
}
````

In this case neither method name nor generic argument gets highlighted correctly (face is `default`). Removing the access-modifiers doesn't seem to make any difference.

Apart from that, it all looks good to me. I'm happy with the code and the structure. If we can get the few bugs fixed too, that would be great, but I'm not going to make it a requirement if it turns out to be difficult.

Some test-cases for the new highlighting would be nice. We already have some tests for font-lock code in `typescript-mode-general-tests.el`, like this:

https://github.com/emacs-typescript/typescript.el/blob/c0f11efdfd693aa55a192c8e4a26fc6fcd2c9f90/typescript-mode-general-tests.el#L306-L320

They should hopefully not be too hard to extend to include some of the new things covered by level 4.

Do you need any help with that?

File: typescript-mode.el [hunk 2]
```diff
@@ -270,20 +270,43 @@ Match group 1 is MUMBLE.")
...
                                  "true" "arguments" "this"))
   "Regular expression matching any future reserved words in typescript.")
 
+(defconst typescript--builtin-re
+  (typescript--regexp-opt-symbol
+   '("console"))
+  "Regular expression matching builtins.")
+
+(defconst typescript--function-call-re "\\(\\w+\\)\s*("
+  "Regular expression matching function calls.")
```
josteink: According to my testing in `font-lock-studio` this regexp also catches simple method-declartions in classes. This is  a good bonus! :smile: 

But from what I can tell, this expression also needs to be somewhat expanded to catch more complex cases with function-calls (and declarations) where Generic type-arguments are provided.

(I'll give it a spin and see what I can come up with)

josteink:
Another thing I uncovered while replaying a few test-cases in `font-lock-studio` is that a lot of highlights seem to be run many times over. For bigger files, this will lead to degraded performance.

I was uncertain at first why that was, but the answer seems fairly simple:

* `typescript--font-lock-keywords-2` is defined as `typescript--font-lock-keywords-1` plus some more things.
* `typescript--font-lock-keywords-3` is defined as `typescript--font-lock-keywords-2` plus some more things. (so far so good)
* `typescript--font-lock-keywords-4` is defined as "some stuff", then `typescript--font-lock-keywords-2`, `typescript--font-lock-keywords-3` and then plus some more things. **This is NOT good.**

The good news it is simple to fix :)

````diff
     ;; highlights that append to previous levels
     ;;
-    ,@typescript--font-lock-keywords-2
     ,@typescript--font-lock-keywords-3
````

With that simple change, there's less redundant code being run during fontification.

**Comment by josteink** [comment]
This other change seems to pick up simple generic function-calls and declarations with 1 generic type-argument:

````diff
   "Regular expression matching builtins.")
 
-(defconst typescript--function-call-re "\\(\\w+\\)\s*("
+(defconst typescript--function-call-re "\\(\\w+\\)\\(<\\w+>\\)?\s*("
   "Regular expression matching function calls.")
````

This will highlight the function-name, but not the type. I still consider this an improvement worth looking at.

**Comment by tam5** [comment]
need some more time for the tests and generics

**Comment by josteink** [comment]
No rush. Take your time. 

I’ll be happy to merge this once final :smile:

**Comment by tam5** [comment]
i can already say, there's a good chance we might have to settle for setting an arbitrary limit on the number of nested generics we would support. I'm talking about:

```typescript
const thing: Map<any, <A<B<C<D>>>>
```

because there is no support for a recursive regexp. It seems pretty unlikely (and probably smelly code anyway) to have say > 5 levels, but still less than ideal. Maybe we can check `java-mode` and see how they are handling that

josteink:
Agreed. This could be solved by recursive regexps, but I don’t think Emacs supports those. 

IMO 1 or 2 levels is fine for now.

josteink:
> Maybe we can check java-mode and see how they are handling that

`java-mode` is derived from `cc-mode` which is a kinda heavy framework for these kind of things.

Skimming the `cc-mode` code I can't find an obvious answer for `java-mode` specifically, but c++ mode just looks for `template <.*>`.

Maybe something like that could work if we make the expression explicitly non-greedy, but still exclude new-lines?

josteink:
You know what? I'm happy with this as it is, and right now I can't *bear* going back to git `master` for regular work.

I'm going to merge this, and if you want to add test-cases, or improve generics-support, feel free to issue more PRs. They will most definitely be welcome 😃

tam5:
👍 sounds good, unfortunately i have 'real' work to do so its taking me some time, but i'll try and follow up when i figure out a good solution

josteink:
This PR seems to be the source for a new fontification bug in level4.

Please see: https://github.com/emacs-typescript/typescript.el/issues/131

PR #111, opened 2019-09-19
better generics support

tam5:
follow up to issue #79 and PR #110.

This should
1) Give better support for highlighting generic types
2) fix function args highlighting when function is defined with generic arg

It does NOT
1) support multiline generic arguments

@josteink I'll either add tests here or to a separate PR. try this out for a bit and see if you spot any bugs

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -283,14 +283,6 @@ Match group 1 is MUMBLE.")
    '("private" "protected" "public" "readonly" "static" "extends" "implements"))
   "Regular expression matching access modifiers.")
 
-(defconst typescript--generic-type-re
-  "<\\(.*\\)>"
-   "Regular expression matching generic types.")
-
-(defconst typescript--generic-type-extended-re
-  (concat "<\\(.*\\)extends\\(.*\\)>")
-   "Regular expression matching generic types with extension.")
-
 (defconst typescript--decorator-re
   (concat "\\(@" typescript--name-re "\\)"))
 
@@ -1878,7 +1870,7 @@ and searches for the next token to be highlighted."
     ;; formal parameters
     ,(list
       (concat
-       "\\_<function\\_>\\(\\s-+" typescript--name-re "\\)?\\s-*(\\s-*"
+       "\\_<function\\_>\\(\\s-+" typescript--name-re "\\)?\\s-*\\(<.*>\\)?\\s-*(\\s-*"
        typescript--name-start-re)
       (list (concat "\\(" typescript--name-re "\\)\\(\\s-*).*\\)?")
             '(backward-char)
@@ -1982,7 +1974,7 @@ context."
     (when (called-interactively-p 'interactive)
       (message "Syntactic context: %s" syntactic-context))
 
-    syntactic-context))
+   syntactic-context))
 
 (defun typescript--class-decl-matcher (limit)
   "Font lock function used by `typescript-mode'.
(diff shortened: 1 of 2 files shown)


josteink:
Great! I'll give it a spin.

Also a heads up: Just to make it easier for all of us, I've invited you to the `typescript-emacs` organization. Welcome to the team!

This should (if I've done things correctly) provide you with PUSH access to the `typescript.el`-repo.

Even though you can in theory push to master, for now it would be nice if you continued using feature-branches, and asked for feedback from other members via PRs like you already do.

As for "your style" of doing commits: IMO they're a bit "chatty" and not exactly "atomic". I'd rather not have commits like that in `master`. If you squash the PRs when merging, that will complete eliminate the problem anyway, so no worries. 👍 

IMO pushing directly to `master` should only be done for completely uncontroversial changes anyway, so don't consider it a restriction on your part. We're all there 😄

josteink:
As for the changes themselves... I see 1 improvement and 1 regression:

image

1. Improved: Additional generic arguments highlighted correctly.
2. Regression: Argument-variables now highlighted as types.

typescript-mode.el
```diff
@@ -2020,13 +2012,18 @@ This performs fontification according to `typescript--class-styles'."
     ;;
     ,@typescript--font-lock-keywords-3
 
+    ;; generics support
+    ,(list
+      (concat "<\\s-*" typescript--name-start-re)
```
josteink: I see no matching `>`for this `<`.

Could that be the reason for the regression I've observed and commented on in the general PR discussion?

tam5: Indeed, that was the reason. I’ve added the fix, as well as added this case to the test

josteink:
Ok so with the recent changes, I've found not found any direct "bugs" yet, but I've found something which ... looks weird.

We're now so good at highlighting generic type-arguments... That returning `Promise`-instances is looking incorrectly highlighted (which is true, since we don't highlight return-types yet):

Screenshot-Sat 21 Sep 2019 07:43:23 PM CEST

`Promise` is a type too, right?

But I guess, we can handle that later if/when we add proper highlighting for return-types (and maybe one day, even function arguments!)

josteink:
I’ve only got nitpicks on this. All in all it looks better than it used to, so I’m happy. 

Are you looking to add more changes, or do you consider this ready to be merged at this point?

tam5:
@josteink what are your nitpicks? I'm happy to hear them.

I'd consider it ready to merge and take an incremental approach to fixing the other issues, but I understand the other point of view which says the more you highlight the more glaring the unhighlighted portions become as you pointed out above. I'll defer to your judgment on that one. The other updates will come, just not exactly sure when.

josteink:
> josteink what are your nitpicks? I'm happy to hear them.

Nothing not already mentioned. No merge-stopped anyway.

Basically my *biggest* complaint is how Promise-heavy APIs now have partial type-highlighting for return-types because of the improved generics-support :smile:

tam5:
@josteink if that's the case, I'd say let's merge this when you're ready please (looks like i don't have write access still), and I'll get started thinking about type highlighting which seems to be one of our last real shortcomings.

josteink:
> (looks like i don't have write access still)

Oh. Really? Let's fix that then! You should have proper push-rights now.

> I'd say let's merge this when you're ready

Nah. I'd rather have you land this one yourself, at least this one time. Wouldn't want to take that away from you :smile:

zhenwenc:
@tam5 What would be the expected highlighting for this JSX? 
```
const FooComponent = () => (
  <div className="bar">
    <p>The answering fool riots beside a horrified tomato.</p>
  </div>
);
```

PR #114, opened 2019-09-24
Update Emacs Git to Xenial-build

josteink:


josteink:
PR just to see CI pass before merging.

PR #116, opened 2019-10-17
Add `delete-selection` hints

elibarzilay:
Makes it properly delete selected text if any.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -764,6 +764,7 @@ comment."
                    (eq (current-column)
                        (1+ (current-indentation)))))
       (indent-according-to-mode))))
+(put 'typescript-insert-and-indent 'delete-selection t)
 
 (defun typescript-insert-and-autoconvert-to-template (key)
   "Run the command bount to KEY, and autoconvert to template if necessary."
@@ -771,6 +772,7 @@ comment."
   (call-interactively (lookup-key (current-global-map) key))
   (when typescript-autoconvert-to-template-flag
     (typescript-autoconvert-to-template)))
+(put 'typescript-insert-and-autoconvert-to-template 'delete-selection t)
 
 ;;; Syntax table and parsing
 


josteink:

Before I move on, I'm must admit I'm not quite sure I follow what problem this solved.

Can you provide a simple example of what you feel doesn't work, and how this patch makes that scenario work better?

elibarzilay:
It makes it play nicely with the [`delete-selection-mode`](https://www.emacswiki.org/emacs/DeleteSelectionMode) feature in Emacs: with this feature, if some text is selected and you type something, then the new text replaces what was selected.  For this to work, some functions that insert text need to be marked using this property to get the same behavior.

To try it, turn the mode on, select some text and type something.  Then, in typescript mode, type a `"` (for example) when text is selected and see that the selection is *not* deleted (and that's what my patch resolves).

josteink:
I’m not sure I can reproduce this error, and I use delete-selection-mode globally. 

What version of Emacs are you using?

josteink:
Can you show a concrete example-buffer or something, and a concrete edit where this will make cause change in behaviour?

It's hard for me to accept a merge if I cannot tell what changes it brings. I hope you understand.

elibarzilay:
I put this configuration in `x.el`:

    (package-initialize)
    (add-hook 'typescript-mode-hook (lambda () (tide-setup)))
    (delete-selection-mode +1)
    (require 'tide)

Then, start Emacs with *just* that:

    emacs -q -l x.el x.ts

And to try it:

    foo <home> <S-end> bar <home> <S-end> (

and note that `bar` properly replaces `foo`, but `(` gets appended instead of replacing the text.

tide

Trying the whole thing with the two properties fixes this.

josteink:
Thanks for the detailed description. Now I can *definitely* reproduce that!

Maybe we can even create some automated tests for it, to prevent regression?

If it's OK with you, I'll take a look at that, and if we can, I think it's worth adding to the PR. Does that sound good to you?

elibarzilay:
I tried a few things, but I don't see any good way to fake input for `typescript-insert-and-indent` (or to `self-insert-command`)...  If you know of any way to do that I'll try it.

josteink:
OK. So looking into this a little more in depth, it seems like the "magic" behind Emacs-selections happens transparently inside functions when they are declare `(interactive "^")`. `interactive` is also a C-function and therefore it's very hard to fake the Shift-selection behaviour in a test-context.

So yeah. Automatically testing this is a lot harder than it initially seemed, and I'm fine with this PR going through without the any additional tests.

And as far as I can see testing them, these changes does what they promises...  So consider it merged!

Congrats on your first typescript PR. Feel free to send us more in the future :)

elibarzilay:
@josteink Actually, that part would be easy -- since it's simple to write code to set and activates the region without relying on the `interactive` bit, so if you figure out a way to test call `typescript-insert-and-indent` from a test I can implement it...

And thanks!

lddubeau:
Has anyone tried `(call-interactively (key-binding some-key-binding))`. That's what [I've used in the past](https://github.com/ananthakumaran/tide/blob/master/tide-tests.el#L249) to simulate keystrokes.

josteink:
I've tried something like the test-case below, but it doesn't work:

````emacs-lisp
;; delete-selection-mode tests

(ert-deftest handles-delete-selection-mode-properly ()

  (with-temp-buffer
    (typescript-mode)
    (delete-selection-mode 1)
    (insert "foo")
    (call-interactively (key-binding "<home>"))
    (call-interactively (key-binding "S-<end>"))
    (insert "bar")
    (call-interactively (key-binding "<home>"))
    (call-interactively (key-binding "S-<end>"))
    (insert "(")

    (should
     (equal "(" (buffer-substring-no-properties (point-min) (point-max))))))
````

From what I can tell `(key-binding "<home>")` resolves til `nil` (same as `"C-a"`)... I would assume this is one of those "always defined" key-bindings, so not those gets resolved in the real world.

I tried being explicit about what functions and keys should be call, with something terrible like this, but still it seems not to apply as it would when done by a "real" user:

````emacs-lisp
(ert-deftest handles-delete-selection-mode-properly ()

  (with-temp-buffer
    (typescript-mode)
    (delete-selection-mode 1)
    (insert "foo")
    (call-interactively #'move-beginning-of-line nil (vector (kbd "<home>")))
    (call-interactively #'move-end-of-line nil (vector (kbd "S-<end>")))
    (call-interactively #'self-insert-command nil (vector (kbd "b")))
    (insert "ar")
    (call-interactively #'move-beginning-of-line nil (vector (kbd "<home>")))
    (call-interactively #'move-end-of-line nil (vector (kbd "S-<end>")))
    (call-interactively #'self-insert-command nil (vector (kbd "(")))

    (should
     ;; resolves to 'fooar'
     (equal "(" (buffer-substring-no-properties (point-min) (point-max))))))
````

Anyone have any good ideas on how we can land this?

**Comment by lddubeau** [comment]
@josteink If I call `(key-binding (kbd "<home>"))` I get the function bound to `<home>`. It's been a very long time since I've looked into the details of the key binding system in Emacs so I don't have a ready explanation why `kbd` is needed.

**Comment by josteink** [comment]
Thanks for the the tip. With that in place, the code becomes, and it kinda runs... Unless it doesn't.

````emacs-lisp
(ert-deftest handles-delete-selection-mode-properly ()

  (with-temp-buffer
    (typescript-mode)
    (delete-selection-mode 1)
    (insert "foo")
    (call-interactively (key-binding (kbd "<home>")))
    (call-interactively (key-binding (kbd "<S-end>")))
    (call-interactively (key-binding (kbd "b")))
    (insert "ar")
    (call-interactively (key-binding (kbd "<home>")))
    (call-interactively (key-binding (kbd "<S-end>")))
    (call-interactively (key-binding (kbd "(")))

    (should
     (equal "(" (buffer-substring-no-properties (point-min) (point-max))))))
````

This runs better, but it fails with the same `nil` error when trying to look up  (key-binding `(kbd "<S-end>"))` in a test-context, despite that (ironically) working fine interactively.

Meh.

**Comment by elibarzilay** [comment]
That won't work because `<S-end>` doesn't have a binding.  Since it doesn't, when you actually use it, you get the binding for `<end>` and *that* command notices that it was actually invoked with shift (via the special `(interactive "^...")` thing).

FWIW, I tried something like this:

```
(ert-deftest handles-delete-selection-mode-properly ()
  (with-temp-buffer
    (typescript-mode)
    (delete-selection-mode 1)
    (insert "foo")
    (set-mark (- (point) 1))
    (activate-mark)
    (call-interactively 'typescript-insert-and-indent t [?\(])
    (should
     (equal "fo(" (buffer-substring-no-properties (point-min) (point-max))))))
```

But it doesn't work -- and I think that the problem is that `(this-command-keys)` doesn't get the fake keys (might be a bug).

PR #119, opened 2019-11-20
force generic regex to include symbol before it

tam5:


josteink:
This change does not include any test-cases, so it’s hard to know what this fixes. 

Also there’s no guarantees that we won’t have future regressions. Could you consider adding a test for this change?

PR #120, opened 2019-11-20
add a test for simple tsx

tam5:
This adds a missing test for #119. 

It also includes a `Dockerfile` and some instructions that can ease testing.

Currently, the case provided in the test is `<div>test</div>`and `div`is not treated as a generic.

However, a case with multiple tags on a  line like `<div>test <span>me</span></div>` would indeed highlight the `span` as a generic type. Still open to solutions for this, although I personally just recommend web-mode for this

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM ubuntu:bionic
+
+RUN apt-get update && \
+  apt-get install -y \
+  make curl python git emacs
+
+RUN curl -fsSL https://raw.githubusercontent.com/cask/cask/master/go | python
+ENV PATH="/root/.cask/bin:${PATH}"
+
+RUN mkdir -p /home/typescript.el
+COPY . /home/typescript.el
+WORKDIR /home/typescript.el
+
+CMD ["make", "test"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN apt-get update && \
+  apt-get install -y \
+  make curl python git emacs
+
+RUN curl -fsSL https://raw.githubusercontent.com/cask/cask/master/go | python
+ENV PATH="/root/.cask/bin:${PATH}"
+
+RUN mkdir -p /home/typescript.el
+COPY . /home/typescript.el
+WORKDIR /home/typescript.el
```
josteink: Super-minor nitpick: If we're running the tests as `root` anyway, it seems a bit strange to create a "user"-folder in `/home` for our content. How about just `/typescript-mode` ?

josteink:
Thanks for following up on my constant nagging for tests 😄 

I only have one tiny nitpick. Consider it a *suggestion*, but not a merge-stopper in any way.

If you don't think it's worthwhile, you can just go ahead and merge 👍

josteink:
LGTM. 🙂

PR #122, opened 2019-12-06
Include symbol constituents in typescript--function-call-re

sebastiansturm:


josteink:
Hey there thanks for the patch!

All contributions are appreciated, but we need to understand what the changes are before we can except and merge them.

Could you please provide an example of a syntax which prior to this patch was not supported, but which this patch as support for?

sebastiansturm:
I'm often using snake_cased function names (even if it may go against TypeScript conventions). So I'd have something like `some_long_function(a,b,c)`, of which only `function` would be highlighted as a function name. Not sure if what I've done is the best way to fix this, but at least I'm currently getting the syntax highlighting I expected.

josteink:
Ok. Sounds reasonable. Could you add a test case for that, so we can ensure it keeps working in the future?

sebastiansturm:
sure. Am at work right now, will do that this evening or tomorrow. Thanks for the quick response!

sebastiansturm:
in hindsight I could have done that right away, didn't know the regression test suite was so convenient to extend. I have now added a snake_cased function call to `font-lock/level-four`

josteink:
Tested and verified. Looks good to me!

sebastiansturm:
thanks!

PR #124, opened 2020-01-10
Smartparens compatibility

### LINKED ISSUE #118: Incompatibility with smartparens (wrap in quotes)
With smartparens you can select text, then write `"` which will wrap the selected word/text in quotes 
 (like so: `"foobar"`). The commit https://github.com/emacs-typescript/typescript.el/commit/f20103a4487a404d11521305db63f550d9eb3fe1 breaks this functionality. The buffer is then totally f**ked up, even undo doesn't work correctly. `typescript-mode` is the only mode in my setup that has this problem. To reproduce this problem installing the newest version of `smartparens 20191015.1754` and `typescript-mode 20191025.1425` from melpa should work. Also `Emacs 26.3`.

Hope you understand what I mean. If you need more info don't hesitate to ask.

Thx for this awesome mode! ❤️

lddubeau:
Closes #118 

Hooking into the `electric-indent-mode` takes care of most problems.

For the quotes, the functionality is reimplemented as a `post-self-insert-hook`. I have some reservations about this because the `post-self-insert-hook` is really Emacs-wide but the auto-conversion is meant to happen only in buffers where `typescript-mode` is the major mode. I've tried a few other approaches but they all ran into issues.

@josteink Opinion about the use of `post-self-insert-hook` here?

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -743,28 +743,11 @@ the string from a plain string to a template."
 
 (defvar typescript-mode-map
   (let ((keymap (make-sparse-keymap)))
-    (dolist (key '("{" "}" "(" ")" ":" ";" ","))
-      (define-key keymap key #'typescript-insert-and-indent))
     (dolist (key '("\"" "\'"))
       (define-key keymap key #'typescript-insert-and-autoconvert-to-template))
     (define-key keymap (kbd "C-c '") #'typescript-convert-to-template)
     keymap)
   "Keymap for `typescript-mode'.")
-
-(defun typescript-insert-and-indent (key)
-  "Run the command bound to KEY, and indent if necessary.
-Indentation does not take place if point is in a string or
-comment."
-  (interactive (list (this-command-keys)))
-  (call-interactively (lookup-key (current-global-map) key))
-  (let ((syntax (save-restriction (widen) (syntax-ppss))))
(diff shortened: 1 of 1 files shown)


josteink:
I’ve never seen it before to be honest, but I also see you’ve guarded the hook to only apply to `typescript-mode`, so I don’t fear it will cause any damage outside our own major-mode. 

The implementation itself looks simple enough, but this change in general, is quite outside my “natural (Emacs) habitat”. 

If you are comfortable with this change, I don’t think I can come up with a reason we shouldn’t move forward with this solution.

lddubeau:
@josteink I just wanted to make sure there wasn't a gotcha I missed, but that you knew about. Lately I've been feeling that there are a lot of gotchas in Emacs, and it is hard to keep them all in mind all the time.

I'm going to wait for the feedback I requested from the folks who experienced the issue and then I'll merge.

josteink:
Sounds like a good approach to me. 

And yeah, while there will always be gotchas, I’d like to hope at least this major-mode has a reasonably hack-free codebase (unlike other major-modes I maintain...)

zeusdeux:
@lddubeau Can confirm it works for me as well. Thanks!

lddubeau:
Feedback was positive. So it is merged.

PR #125, opened 2020-02-01
fix(font-lock): prefer fontifying constants over keywords

Fuco1:
This way null, false, undefined etc. will be fontified with
font-lock-constant-face instead of font-lock-keyword-face.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -294,13 +294,13 @@ Match group 1 is MUMBLE.")
 
 (defconst typescript--font-lock-keywords-2
   (append typescript--font-lock-keywords-1
-          (list (list typescript--keyword-re 1 font-lock-keyword-face)
+          (list (cons typescript--constant-re font-lock-constant-face)
+                (list typescript--keyword-re 1 font-lock-keyword-face)
                 (list "\\_<for\\_>"
                       "\\s-+\\(each\\)\\_>" nil nil
                       (list 1 'font-lock-keyword-face))
                 (cons "\\_<yield\\(\\*\\|\\_>\\)" 'font-lock-keyword-face)
-                (cons typescript--basic-type-re font-lock-type-face)
-                (cons typescript--constant-re font-lock-constant-face)))
+                (cons typescript--basic-type-re font-lock-type-face)))
   "Level two font lock keywords for `typescript-mode'.")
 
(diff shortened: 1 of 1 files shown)


josteink:
Hey @Fuco1. 

Sorry about the slow reply here, but I’m currently really busy with a few things at work. 

I promise to review your changes first thing when I have a spare minute 🙂

Fuco1:
No worries, take your time :) I currently run my own branch in my config so I'm fine :D

Thanks!

josteink:
Hey @Fuco1.

I'm looking over this change now, and trying to assess what scenarios (code-examples) this change would impact.

The *absolutely* best way to provide this, would be in the form of a test, so that we ensure that the new behaviour this change introduces stays fixed unless we intend to change it.

But if you don't have time for that right now, at least a piece of sample code would be nice to look at :smile:

Fuco1:
Basically `true`, `null`, `false` and `this` will be fontified as constants and not keywords (like `if`, `else` etc).

It's only these four:

```emacs-lisp
(-intersection
 ;; typescript--constant-re
 '("false" "null" "undefined" 
   "Infinity" "NaN"
   "true" "arguments" "this")
 ;; typescript--keyword-re
 '("abstract" "any" "as" "async" "await" "boolean" "bigint" "break" "case" "catch" "class" "const"
   "constructor" "continue" "declare" "default" "delete" "do" "else"
   "enum" "export" "extends" "extern" "false" "finally" "for"
   "function" "from" "get" "goto" "if" "implements" "import" "in" "instanceof"
   "interface" "keyof" "let" "module" "namespace" "never" "new" "null" "number" "object" "of"
   "private" "protected" "public" "readonly" "return" "set" "static" "string"
   "super" "switch"  "this" "throw" "true"
   "try" "type" "typeof" "unknown" "var" "void"
   "while")) ;; => ("false" "null" "true" "this")
```

Fuco1:
I can add a test, I'll have a look at how you do testing.

josteink:
Sounds great. Let me know if you need help. 

Use `make test` to get started 🙂

josteink:
If there's no tests, I still agree this is a proper thing to do.

Another approach might be to remove those things as keywords... But for now, I consider this good enough :)

PR #127, opened 2020-02-28
Fix indentation infinite looping

Jevri:
With unbalanced parentheses typescript--backward-to-parameter-list goes in to an infinite loop.

### DIFF
--- a/test-files/indentation-reference-document.ts
+++ b/test-files/indentation-reference-document.ts
@@ -724,3 +724,7 @@ const blip999 = (window.location.href === "fnord" ?
 if (true) {
     .bleh();
 }
+
+) {
+    1 + 2; // Indenting this line would cause an infinite loop
+}
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2432,7 +2432,7 @@ moved on success."
                                  (looking-at "\\_<\\(switch\\|if\\|while\\|until\\|for\\)\\_>\\(?:\\s-\\|\n\\)*(")))))
                     (condition-case nil
                         (backward-sexp)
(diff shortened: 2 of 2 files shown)


lddubeau:
Thank you for the PR. It needs a test case that fails when the patch is not applied. By default we add such cases to `indentation-reference-document.ts` but I'm thinking that you've run into a case that most likely to be triggered if the parameters that should be iterated over at at the start of the file,  with no other code prior to the function that has unbalanced parentheses. If that is the case, then a new file can be added for this test and should contain a comment mentioning that this test file should not be edited to add new cases.

josteink:
This PR breaks the CI, so I checked the log:

> FAILED  indentation-reference-document-is-reflowed-correctly

Basically this indentation fix breaks other indentation rules. If you run this test manually, you may find out which aspect of the reference-document is reflowed/indented incorrectly and take it from there.

Jevri:
I can see the failure comes from the test I just added. I am not sure what should be done here, on line 2481 we assume it is safe to call (backward-list) if point is after a ')', but in this case it fails as it has no match. Should we have landed in the default case on line 1487 or should point have been in a different place all together from the beginning?

josteink:
Hey @Jevri.

Thanks for the update! I'm looking at reasonably simple change of code, *and* I see a simple test-case for it.

That makes me happy about this PR and unless anyone else has any objections, I'll merge this within a few days :+1:

josteink:
Merged it is! Thanks for helping out!

Feel free to send in more patches in the future if you see anything worth fixing :+1:

Jevri:
@josteink Great, I sure will if I find anything :)

PR #134, opened 2020-04-03
Make keywords customizable.

fredefox:


fredefox:
Caveat: I haven't actually tested this locally. I'm not a super pro elisp hacker. One thing I'm unsure about is the interaction with `defcustom` and `defconst`. If I customize it, will it be picked up when `typescript--keyword-re` is defined?

My use-case is that I don't want `get` and `set` to be highlighted as keywords.

josteink:
This does in my opinion seem like a strange thing to customize, but ... I guess I only speak for myself in that regard.

> Caveat: I haven't actually tested this locally

That's an absolute minimum thing you have to do before issuing a PR and expecting any real feedback on it.

Can you do that, and let us know when you've verified that this works as intended?

fredefox:
> This does in my opinion seem like a strange thing to customize
> That's an absolute minimum thing you have to do before issuing a PR and expecting any real feedback on it.
Feel free to dismiss my submission if you find it so problematic.

fredefox:
FWIW the EcmaScript specification doesn't consider either of `arguments`, `get` or `set` as keywords. The former of these 3 are considered a constant by this library.

MarcusRiemer:
I agree with both of you: Handing over untested code can be a burden for a project, so tests are important. But the current state does indeed highlight some tokens as keywords that aren't: I work quite a bit with RxJS, where `of` is a often recurring function name which is also (incorrectly) highlighted by this mode. For me its more a subtle annoyance and not something that really bothers me.

But what about just dropping the tokens that are not proper keywords?

lddubeau:
I'm also in the camp of finding it a "subtle annoyance".

> But what about just dropping the tokens that are not proper keywords?

 JS' grammar allows them to be keywords sometimes, depending on context. (AFAIK, TS merely follows JS.) From [the spec](https://tc39.es/ecma262/#sec-keywords-and-reserved-words):

> Those that are always allowed as identifiers, but also appear as keywords within certain syntactic productions, at places where Identifier is not allowed: as, async, from, get, of, set, and target.

Whether they are identifiers or keywords depends on context. The mode currently always treats them as keywords, which is incorrect. Always treating them as identifiers is also incorrect. I'm 100% certain that if we replace one incorrect behavior with another, we'll have folks complain about the new behavior. They're going to be as justified in their complaint as people who complain now that the behavior is incorrect.

I see two options:

1. Fixing highlighting to highlight these "sometimes-keywords" correctly in all cases. I expect getting this right to be Damn Hard (tm).

2. Using a configuration option to let users decide which wrong highlighting they prefer. We're wary of adding too many configuration options but this may be the pragmatic answer here. This change needs to be brought forth by someone who is actually bothered enough by the current behavior (that's not me) to be willing to champion this change and test it.

fredefox:
Personally I would prefer them to be highlighted as identifiers. I'm sorry if my contribution doesn't live up to the standard that you expect. Unfortunately I lack the required skillset to meet those expectations. So I'm not going to follow up on this. I apologise for any inconvenience I may have caused you.

josteink:
> I agree with both of you: Handing over untested code can be a burden for a project, so tests are important. 

I honestly didn’t asks for automated unit-tests tests. 

I just asked for the code to be *actually tested* before anyone should be asked to review a patch. 

OP admitted to not having conducted any such testing, hence my objection.

tam5:
@fredefox don't give up!

Why don't you give us some information about how you currently manage packages, and we can help you set yourself up so that you can run your code locally and test it out.

PR #136, opened 2020-08-03
Fix duplicate test name

phst:


josteink:
Bloody nice catch!

How oh how did you discover this one? 😀

PR #137, opened 2020-08-08
Add option for list items indenting

sirikid:
This option is for people who like formatting like this:
```ts
const x =
    [ foo
    , bar
    , baz
    ]
```


josteink:
Change looks harmless enough, but I'd like to see some test-cases added for it before merging.

Feel free to create an [alternate indentation reference document](https://github.com/emacs-typescript/typescript.el/tree/master/test-files) for you mode, and then add a similar ert-testcase to [the ones we already have](https://github.com/emacs-typescript/typescript.el/blob/master/typescript-mode-general-tests.el#L21-L53).

Let us know if you need help, but it shouldn't really be that hard. :+1:

sirikid:
@josteink done.

josteink:
That looks great. Thanks!

Consider it merged!

PR #139, opened 2020-09-01
Use ‘(should-not …)’ instead of ‘(should (not …))’.

phst:
This is equivalent, but shorter, and produces better error messages.


josteink:
Fully agreed. LGTM. 

Thanks for the patch. Consider it merged 😃

PR #140, opened 2020-09-01
Use ‘insert-file-contents’ instead of ‘find-file’ for tests.

phst:
‘find-file’ doesn’t signal errors if visiting the file failed and is therefore
more brittle.  Also it makes the file buffer readonly if the visited file is
readonly, causing tests to fail if the testfiles are readonly.

### DIFF
--- a/typescript-mode-general-tests.el
+++ b/typescript-mode-general-tests.el
@@ -19,7 +19,8 @@
   (untabify (point-min) (point-max)))
 
 (ert-deftest indentation-reference-document-is-reflowed-correctly ()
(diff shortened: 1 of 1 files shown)


josteink:
Unreliable tests are the worst, so thanks for helping out.

Looking at the code, it does seem a bit repetitive (`(with-temp-buffer (insert-file-contents) ....)`), but on the flip side, we don't have to do manual book-keeping/cleankup (`(kill-buffer buffer)`).

I guess you can consider it merged. Thanks for contributing! :)

PR #141, opened 2020-09-29
* typescript-mode.el: Use EOS instead of EOL for auto mode regexp.

cireu:
This prevent matching filename like "xxxx.ts\nxxxx". A corner case which satisfies POSIX filename standard but weird for daily use.

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2978,7 +2978,7 @@ Key bindings:
      (folding-add-to-marks-list 'typescript-mode "// {{{" "// }}}" )))
 
 ;;;###autoload
-(add-to-list 'auto-mode-alist '("\\.ts$" . typescript-mode))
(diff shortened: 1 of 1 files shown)


josteink:
Can I ask how you ended up actually finding and fixing a bug like this? 😅

cireu:
I found many regexp in `auto-mode-alist` are ended with `\\'` but `typescript-mode` use `$`. So curiosity drives me to find the answer.

You can create a file which name contains a linebreak by pressing `C-j` in `find-file` prompt.

cireu:
Do you have any other questions? @josteink

josteink:
Thanks for the answer, and sorry about the slow reply. I can confirm your findings, and that absolutely makes sense.

Consider this merged :)

PR #152, opened 2021-09-20
fix eval invocation in typescript--re-search-backward to work with lexical scoping

Rogach:
Under lexical scoping variable references won't be found when evaluating the expression, resulting in "eval: Symbol’s value as variable is void: regexp" error.

By the way, do we really need the `eval` here?

### DIFF
--- a/typescript-mode-lexical-binding-tests.el
+++ b/typescript-mode-lexical-binding-tests.el
@@ -1,0 +1,25 @@
+;; -*- lexical-binding: t -*-
+;;; typescript-mode-lexical-binding-tests --- This file contains test for typescript-mode.el under enabled lexical-binding
+
+;;; Commentary:
+;; To know how to run the tests, see typescript-mode-tests.el
+
+(require 'ert)
+
+;; reload typescript-mode with lexical-binding enabled
+(with-temp-buffer
+  (insert-file-contents "typescript-mode.el")
+  (ignore-errors
+    (while (setq sexp (read (current-buffer)))
+      (eval sexp t))))
+
+(require 'typescript-mode-test-utilities)
+
+(ert-deftest lexical-binding--indentation-does-not-throw-error ()
+  (with-temp-buffer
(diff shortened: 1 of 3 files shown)


josteink (approved):
Not sure about the need for `(eval)` but apart from the the change itself looks good to me!

josteink:
While the change itself looks good, it reflects on something which used to be broken while the build was still “green”.  

Would it be possible to add a test-case for this to prevent future regressions?

Rogach:
Yes, I'll add the test tomorrow (should be easy - just explicitly set lexical-binding to non-nil).

Rogach:
Hm, no, seems that it's not that easy - you can't change `lexical-binding` value during execution, it's only changed on per-file basis or globally.

I asked over in emacs irc and one suggestion was "write a temp file with the setting and code and eval it", but that seems way to involved for a simple test.

One alternative is to switch all tests&code to lexical scoping (seems that lexical scoping is the main way forward nowdays).

josteink:
You could always create a dedicated file for tests in lexical mode. We already have different test files for different specific sub-genres of tests, so this would really not be any different.

josteink:
Also I have fixed the CI so that all the builds are green now, so if you merge in latest changes from master, this PR should be all green as well 🙂

Rogach:
I've added the test, but it's horrible (I had to re-read and re-evaluate typescript-mode.el to get it to load with lexical-binding enabled). I'm not even sure how I managed to stumble upon the original bug - it seems that lexical-binding can be enabled only from inside the library file. It's possible that it was some regression inside Emacs itself - I'm running builds from git master, and I can't reproduce my original bug under the latest build.

> Also I have fixed the CI so that all the builds are green now, so if you merge in latest changes from master, this PR should be all green as well

I rebased the changes on top of latest master, hopefully the CI runs fine =)

josteink:
Lol. That’s absolutely terrible indeed, but it does the job. It also doubles as a preparation for migrating to lexical mode proper. 

Consider it merged! Thanks for the patch!

PR #159, opened 2021-11-21
Fix cl deprecation warning

tancnle:
This commit replaces `cl `with `cl-lib` to remove deprecation warning on Emacs 27 and onward (see deprecation [note](https://github.com/emacs-mirror/emacs/blob/8758c96dc7f252944eb97b95c92081b157084477/etc/NEWS#L601)).

## Steps to reproduce
- Install `typescript`
- Run Emacs v27 (e.g. 27.2)
- Observe the deprecation warning

```
Package cl is deprecated
```

### DIFF
--- a/typescript-mode-general-tests.el
+++ b/typescript-mode-general-tests.el
@@ -7,7 +7,7 @@
 
 (require 'ert)
 (require 'typescript-mode)
-(require 'cl)
+(require 'cl-lib)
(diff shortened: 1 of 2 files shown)


josteink:

Change looks harmless enough. If the CI says it’s ok, I guess this is good to merge 🙂

PR #160, opened 2022-02-19
capitalize mode name "TypeScript"

trev-dev:
I realize this is picky, but TypeScript is a name, and like names (and titles), they should be capitalized. Thanks for the light-weight major mode!

Screenshot from 2022-02-18 16-26-30

vs

Screenshot from 2022-02-18 16-27-05

### DIFF
--- a/typescript-mode.el
+++ b/typescript-mode.el
@@ -2911,7 +2911,7 @@ the broken-down class name of the item to insert."
 ;;; Main Function
 
 ;;;###autoload
-(define-derived-mode typescript-mode prog-mode "typescript"
+(define-derived-mode typescript-mode prog-mode "TypeScript"
   "Major mode for editing typescript.
 
 Key bindings:


josteink:
Hey hey! Thanks for the PR!

Just one question: Does this impact language detection in other modes like lsp-mode or dap-mode?

Have you tested that? 😄

trev-dev:
Hmm, it didn't yesterday after I changed it, but I run Emacs as a daemon. After rebooting and firing up for some super healthy Saturday morning work, all LSP would look for is eslint...

Leave this with me and I'll see what I can do

trev-dev:
Hokay so I double-checked the documentation on `defined-derived-mode` and the 3rd argument is indeed just a descriptor for the mode, I don't believe it has any functional impact other than the modeline display.

Also, I have this fun thing that happens with lsp-mode where it sometimes fails at finding the correct project root and I have to delete my workspace folder and re-try. I did this and LSP-Mode found my `tsconfig` and fired up.

I did not test `dap-mode` as I do not use it :)

josteink:
If so, let's just merge it.

If it causes trouble down the line, surely somebody will let us know :)

PR #161, opened 2022-02-23
Update build.yml

josteink:
Let more of the build complete even though some things fail.


josteink:
This PR seems to make the build green*er*, so merging it as is, because it at least seems to represent an improvement.

PR #162, opened 2022-04-15
Fix CI test

jcs090218:
I have recently created a tool named [Eask](https://github.com/emacs-eask/eask). It's similar to Cask, but it has a better cross-platform capability. It tests your package/configuration in a much more consistent way.

This patch does the following:

1. Replace Cask with Eask
2. Add test for Emacs 28.1
3. `build.yml` now has only one job (merged unix + windows)
4. Fix test in `windows-latest`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,14 +9,14 @@ on:
   workflow_dispatch:
 
 jobs:
-  unix-build:
+  build:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         experimental: [false]
-        emacs: [26.1, 26.2, 26.3, 27.1, 27.2, 28.1]
+        emacs_version: [26.1, 26.2, 26.3, 27.1, 27.2, 28.1]
         include:
         - emacs: snapshot
           experimental: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,13 +18,13 @@ jobs:
         experimental: [false]
         emacs-version: [26.1, 26.2, 26.3, 27.1, 27.2, 28.1]
         include:
-        - emacs: snapshot
+        - emacs-version: snapshot
           experimental: true
           os: ubuntu-latest
-        - emacs: snapshot
+        - emacs-version: snapshot
           experimental: true
           os: macos-latest
-        - emacs: snapshot
+        - emacs-version: snapshot
           experimental: true
           os: windows-latest
     continue-on-error: ${{ matrix.experimental }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,7 +45,7 @@ jobs:
 
       - uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '16'
 
       - uses: emacs-eask/setup-eask@master
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,7 +16,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         experimental: [false]
-        emacs_version: [26.1, 26.2, 26.3, 27.1, 27.2, 28.1]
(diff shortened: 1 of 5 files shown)


jcs090218:
@josteink Can you take a look at this PR as well? Only when you have time! :)

.github/workflows/build.yml
```diff
@@ -9,71 +9,49 @@ on:
...
-          python-version: "3.6"
-          architecture: "x64"
+          version: ${{ matrix.emacs-version }}
 
-      - uses: purcell/setup-emacs@master
+      - uses: jcs090218/setup-emacs-windows@master
+        if: matrix.os == 'windows-latest'
         with:
-          version: ${{ matrix.emacs }}
+          version: ${{ matrix.emacs-version }}
```
josteink, line 44: Not that I'm eager for an impenetrable layer of sub-tasks, but could this dispatch be delegated into its own single task which uses this approach to create a single, working cross-platform bootstrapping-task, which again depends on these two different setup-tasks?

Or does GitHub actions not support such a construction? 🙂

josteink: I'm thinking something like:

    - uses: jcs/setup-emacs-xplat
      - with:
        os: matrix.os
        version: matrix.emacs-version

.github/workflows/build.yml
```diff
@@ -9,71 +9,49 @@ on:
...
-  #         version: ${{ matrix.emacs }}
-
-  #     - uses: conao3/setup-cask@master
-  #       with:
-  #         version: 0.8.4
-
-  #     - name: Run a multi-line script
-  #       run: |
-  #         emacs --version
-  #         make test
```
josteink, line 79: Delete, rather than comment out. We have version-control 😉

josteink (commented):
On the phone so can't do full review now. I'll check back later.

.github/workflows/build.yml
jcs090218: I guess those are deleted. :)

.github/workflows/build.yml
jcs090218: I am using `if:` statement,

```yml
      - uses: jcs090218/setup-emacs-windows@master
        if: matrix.os == 'windows-latest'
        with:
          version: ${{ matrix.emacs-version }}
```

Would this be sufficient?

jcs090218:
You can checkout my branch and see the full implementation https://github.com/jcs-PR/typescript.el/tree/ci. Or we can wait until you can use your PC. ;)

.github/workflows/build.yml
josteink: I was thinking it would be nice if we had a single task which could bootstrap emacs for any platform based on a parameter, instead of repeating the same if's and else's in all the GitHub actions definition YML-files.

Such that the **whole** process (from a single package perspective becomes):

      - uses: jcs090218/setup-emacs-xplat
        - with:
          os: matrix.os
          version: matrix.emacs-version

Rather than:

      - uses: purcell/setup-emacs@master
        if: matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest'
        with:
          python-version: "3.6"
          architecture: "x64"
          version: ${{ matrix.emacs-version }}

      - uses: jcs090218/setup-emacs-windows@master
        if: matrix.os == 'windows-latest'
        with:
          version: ${{ matrix.emacs-version }}

That said, if it's hard to do, it's not that big a deal. Just trying to avoid needless repetition that's all :)

jcs090218: Unfortunately, this is currently impossible in the Yaml file. We would need to contribute to the action repository (either `setup-emacs-windows` or `setup-emacs`) to have them support other missing platforms.

I might create a new action in the near future by merging the two actions into one, so then we could clean this up! :)

.github/workflows/build.yml
```diff
@@ -9,71 +9,49 @@ on:
...
+        if: matrix.os == 'windows-latest'
         with:
-          version: ${{ matrix.emacs }}
+          version: ${{ matrix.emacs-version }}
 
-      - uses: conao3/setup-cask@master
+      - uses: actions/setup-node@v2
         with:
-          version: 0.8.4
+          node-version: '14'
```
josteink, line 48: Same comment about Node here. Might as well settle for the latest LTS version 🙂

PR #163, opened 2022-04-21
Use merged setup-emacs action

jcs090218:
From the previous comment in #162, I have created the [jcs090218/setup-emacs](https://github.com/jcs090218/setup-emacs). It should automatically pick up the right Emacs to install on each platform.


josteink:
Nice! Merged! 🤠

PR #165, opened 2022-06-16
added: infer keyword

Ruzie:
In ts `infer` keyword is being used to describe ts compiler about that all types declared by user are explicit, in conditional way or returntype
https://www.typescriptlang.org/docs/handbook/2/conditional-types.html


josteink:
Thanks. Looks solid to me. 

Consider it merged 🙂

PR #167, opened 2022-06-30
feat(fontlock): fontify the variable and type in typeguard

Fuco1:
Fontify the variable and type in a type guard.

Before 

image

After

image

### DIFF
--- a/typescript-mode-general-tests.el
+++ b/typescript-mode-general-tests.el
@@ -506,6 +506,19 @@ function foo<Z, Y, Z & Y, Z | Y | Z, Y<X<X, Y>>>()\n"
       "type Thing = number;"
     (should (eq (get-face-at "Thing") 'font-lock-type-face))))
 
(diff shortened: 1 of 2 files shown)


josteink:
Without delving too deep into the code, it looks like a useful improvement. 

Would you be able to add tests for it, so once committed, we know it will keep working? 🙂

Fuco1:
Now looking at this, maybe the `is` should be fontified as keyword as well?

This is from VS Code where the `is` has the same face as the `function` (so keyword)

image

Fuco1:
I will add tests, no problem.

josteink:
Sounds great. Fontifying “is” makes perfectly sense to me.

Fuco1:
You have really neat testing setup for font-locking.  I'm loving the macros :)

josteink (approved):

josteink:
Looks great. Thanks for this PR :)

PR #168, opened 2022-06-30
fix(fontlock): fontify parameters in multiline arglist

Fuco1:
Fix the anchored matcher to allow for multi-line arg list.  I will add tests if we decide to go forward with this change.

Before 
image

after 

image


josteink:
Hey. Seems like we’re still missing some test-cases for the new syntax highlighting you’re adding support for here. 

Let me know when you think this is ready for review!

Fuco1:
I added more test cases and also removed the `continued formal parameter list` rule because it is now covered by the other anchored matcher.  All existing tests pass without it as well and visually I can't tell a difference in the buffer.

josteink:
Simplifications are always welcome. And especially with more tests. Consider it merged!

PR #170, opened 2022-07-08
feat(fontlock): fontify arrow fn arguments

Fuco1:
Fontify the variables in arrow function argument lists.  

Probably won't work 100% because parsing with regexps is hard but 80/20 applies.

Before

image

After

image


Fuco1:
I got around to do class methods as well

image

Also made sure keywords are still fontified as variables when in variable context

image

josteink:
This change is above my paygrade to review, but from what I can tell you've added test-cases and none of the existing one are failing.

Looks good to me, I guess? :laughing:

PR #171, opened 2022-07-09
fix(fontlock): fontify return type in a function type definition

Fuco1:
Fontify as type the `SomeType` in the following:

```typescript
const f: () => SomeType = () => {};
```

Before

image

After

image


josteink:
Looks good to me!

PR #172, opened 2022-07-09
fix(fontlock): do not fontify function calls as keywords

### LINKED ISSUE #103: Function keyword highlight incorrect.
If the function name the same as the keyword, the function also get highlighted.

### Screenshot ###
ts-mode-bug

Fuco1:
If a method/function is named same as keyword or built-in, it should
still be fontified as method/function call.

Before

image

After

image

Fixes #103


josteink:
Tested and looks good to me! Thanks for another fine PR :smile:

PR #173, opened 2022-07-09
fix(fontlock): do not fontify builtins in object/interface key context

Fuco1:
Before

image

After

image


josteink:
I see you’re putting in lots of effort to improve typescript-mode, which is great!

Unfortunately I’m on vacation and won’t be able to review them for the next 2-3 weeks. 

So don’t count lack if feedback as disinterest, but I don’t have a laptop with me, and have more important things to do right now 😉

josteink:
Tested and looks good to me. Consider it merged!

PR #175, opened 2022-08-05
fix(fontlock): only re-fontify method calls with reserved words

### LINKED ISSUE #174: Incorrect highlighting of keywords on latest commit
Keywords such as `if` or `for` are highlighted as functions when standalone or  in comments.

Test the following:
```typescript
if (true) { // if highlights as a function
}

for (const x in y) { // for highlights as a function
}

// same with the below
// if (true) {  
// }

// for (const x in y) {
// }

// foo bar object
```

For me the word `object` is highlighted as a function when in a comment.

Possibly caused by #172

Fuco1:
Functions can not be called after a reserved word such as `if` or
`delete`, however methods can.  Only re-apply fontification on
methods.

Fixes #174

### DIFF
--- a/typescript-mode-general-tests.el
+++ b/typescript-mode-general-tests.el
@@ -397,15 +397,19 @@ snake_cased_function(1, 2, 3)"
       (("<" ">" ",") . nil))))
 
 (ert-deftest font-lock/method-call-with-keyword-name ()
-  "If the name of the function/method is a keyword, it should still be highlighted as function."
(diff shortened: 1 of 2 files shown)


josteink (approved):

josteink:
Thanks! It has better test coverage and build is still green. 

Consider it merged!

Fuco1:
For the future reference: this patch basically undid my previous "fix" and only applied it on methods, as it makes no sense on functions since those are protected by typescript itself (so nobody would write such code).

PR #176, opened 2022-08-08
fix(fontlock): fix special forms/ternary mistaken for method declarations

Fuco1:
Sometimes a special form such as

    if (...) {..}

or a ternary

    a ? b(...) : c

might look like a method declaration and then the insides of the
parentheses get fontified as variables.  Here we add a heuristic to
discard these cases.  It's not 100% but works in most real-life
situations.

Before: 

image

After: (they are fontified normally :D)

### DIFF
--- a/typescript-mode-general-tests.el
+++ b/typescript-mode-general-tests.el
@@ -846,6 +846,22 @@ bbb: Bar,
     (should (eq (get-face-at "aaa") 'font-lock-variable-name-face))
     (should (eq (get-face-at "bbb") 'font-lock-variable-name-face))))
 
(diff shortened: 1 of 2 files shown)


Fuco1:
@josteink friendly ping :)

josteink:
Unit-tests, code comments, documentation of HACK…

How can I resist such a proper PR? Consider it merged 🙂

josteink:
PS: you’ve heard about the tree-sitter based rewrite where all these hacks are no longer needed, right? 😅

Fuco1:
Yea I'm aware, maybe I should help out with that. But for now, this is good :D

PR #179, opened 2022-11-26
Update note after ts-mode rename

alternateved:
Hello, 

It seems `ts-mode` has been renamed in the core to `typescript-ts-mode` as can be seen in [this commit](https://git.savannah.gnu.org/cgit/emacs.git/commit/?id=7fc0eae28f03601e25e2a60030ae2dc59085c6d2) and as was discussed [here](https://lists.gnu.org/archive/html/emacs-devel/2022-11/msg01607.html).

Best regards
alternateved

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ than keeping this old version around.
 
 ## Good news!
 Emacs 29 will ship with tree-sitter support, and it will actually have in-tree
-support for TypeScript!  So now you can just use the provided `ts-mode` and get
+support for TypeScript!  So now you can just use the provided `typescript-ts-mode` and get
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

josteink:
Eh. We may want to revert this one.

Seems the rename is not happening? :smile: 

https://lists.gnu.org/archive/html/emacs-devel/2022-11/msg01651.html

alternateved:
@josteink This is an answer in regards to renaming all `*-ts-mode` to `*-treesit-mode`. Rename from `ts-mode` to `typescript-ts-mode` already happened.

josteink:
Got it. We leave it be then :smile:

PR #182, opened 2023-01-15
Add .tsx to auto-mode-alist

jdufresne:
Add out-of-the-box support for using typescript-mode on *.tsx files. Avoids repetitive manual configuration for end users.


josteink (approved):

josteink:
I think these days one is served better by going for typescript-ts-mode and tsx-ts-mode in Emacs 29, but for those who that is not an option I guess this is OK. 

Just don’t expect anyone to implement “real” TSX-support in typescript.el 😋

PR #195, opened 2025-11-27
Add TypeScript comint REPL integration

tninja:
The code came from https://github.com/emacs-typescript/ts-comint/tree/master

@josteink PTAL Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,31 @@ to `false`: `tsc --pretty false`. However, doing this does more than
 just turning off the colors. It also causes `tsc` to produce less
 elaborate error messages.
 
+## Using `typescript-comint`
+
+`typescript-comint.el` ships with this repository and exposes a simple REPL that integrates with `typescript-mode`.
+
+1. Ensure `typescript-comint.el` is on your `load-path` (it is if you install the package normally).
+2. Start a REPL with `M-x typescript-run-repl`. Use `C-u M-x typescript-run-repl` to edit the command line if you need a different interpreter than the default `tsun`.
+3. From a `typescript-mode` buffer, send code to the REPL using:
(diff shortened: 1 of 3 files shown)


josteink:

Not to be dismissive, but I'm not sure we want to merge this PR. For several reasons.

1. this is kind of a "legacy" typescript-mode now that Emacs core has shipped with a vastly superior tree-sitter based major-mode. Nobody is really working on maintaining this package anymore. Therefore adding more stuff here to maintain is kinda going the wrong direction.
2. you're merging what is already a stand-alone package which can easily be installed alongside whatever typescript-mode you prefer. This one, or the better one, shipped with newer Emacs-versions.
3. merging this here would create code-duplication. If anyone were to provide a fix for a bug/error... Which repo should they contribute to?

As such, I don't feel like merging this unless anyone can come up with some very convincing reasons.

Anyone got any compelling arguments? 😄

tninja:
Got it, yes confirmed that there is built-in typescript-ts-mode with emacs. But looks like there is only 649 lines of code, much less than this package (3k+ lines of code). If typescript-ts-mode is more preferred, I'll use that one.

For the comint part, my experience on other language is that the comint session is critical for a language. So I opened this PR to integrate it in typescript.el, people can use one package (and same style of code), to have editing support and comint. Actually, typescript-ts-mode do not have comint support right now.

I'll close this PR if typescript-ts-mode is the right way to use typescript inside emacs.

josteink:
Its smaller because the parser-code is externalized to tree-sitter. 

It also supports tsx-files through tsx-ts-mode. 

I'm the maintainer of both those new modes too, and I honestly couldn't imagine going back. 

Give it a try!

